PR #27, opened 2016-01-15
Translations.

jason-p-pickering:
Merge translations from translation server.


Markionium:
Merged and published to mvn

PR #300, opened 2017-01-23
Updated inline help mapping links

ceciliapersson:
Updated links as agreed with Mark.

### DIFF
--- a/src/config/inlinehelp-mapping.json
+++ b/src/config/inlinehelp-mapping.json
@@ -75,10 +75,10 @@
     "/edit/validationSection/validationRule/add": "/en/user/html/create_validation_rule.html",
     
     "/list/validationSection/validationRuleGroup": "/en/user/html/about_validation_rule.html",
-    "/edit/validationSection/validationRuleGroup/add": "/en/user/html/create_validation_rule_group.html"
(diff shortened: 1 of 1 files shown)


ceciliapersson:
@nicolayr and @Markionium Do yo know why the check failed? There're no conflicts with the base branch.

nicolayr:
There are some commas missing in the mappings file.

GitHub has syntax highlighting for JSON files, so if you look at the changes that were made to the file it's pretty easy to spot the errors (click the "files changed" tab at the top of this page, which takes you [here](https://github.com/dhis2/maintenance-app/pull/300/files#diff-a8e4921ba0a6e92fbdd6ba5e33a7c219)).

It's a bit misleading however because the error shows up at the ":", while the problem is actually that there's a comma missing at the end of the previous line.

Each line should be the following format:

```
   "section": "help_url",
```

With the exception that the _last_ line in the file should _not_  have a comma at the end.

nicolayr:
That said, this is obviously a very simple fix and it's not a problem for us to do that for you. I just thought I would let you know what the problem was in case you're curious ;)

ceciliapersson:
@nicolayr Thanks for the explanation! I'll be more careful with the commas next time. Thanks for fixing :)

PR #303, opened 2017-02-02
Updated order of "edit" mapping items

ceciliapersson:
Reviewed items and updated order of "edit" mapping items. Also changed all list mappings to the more condensed format.
Could you verify that these two items will render correctly with regards to order and length of item names? 
- "edit" organisationUnitGroup
- "edit" organisationUnitLevel


nicolayr:
Looks fine @ceciliapersson - I pushed a new version now so you can test it on dev/demo soon :)

ceciliapersson:
Thanks @nicolayr!

PR #310, opened 2017-03-08
Properties cannot contain unescaped spaces in keys.

jason-p-pickering:


jason-p-pickering:
@Markionium this is the first part, altering the offending property keys. If you can fix this in the source code, that would be great.

PR #316, opened 2017-04-20
Determine columns and filters for lists by object type

nicolayr:
- Show different columns for different lists
- Add support for filtering lists by object properties


nicolayr:
Already included in #321

PR #348, opened 2017-08-31
Eslinting self fixing

jenniferarnesen:
Ran the eslint with the --fix flag. You might want to use the w=1 url param to filter out the whitespace fixes when viewing the PR.

Error types fixed include:
trailing commas (add or remove)
parentheses in arrow functions
quote style
string templates
else { if  -> else if
prefer const
object shorthand
whitespace
prefer-spread


jenniferarnesen:
There are a few files where I was unsure if the change would cause a problem: 

programRuleConditionField.component.js
eventProgramStore.spec.js
programRuleActionList.component.js
GreyFieldDialog.component.js

nicolayr:
This applies all possible automatic eslint fixes, but there are still 2300 remaining issues.. ;)

PR #351, opened 2017-09-05
Lint text-editor-field and text-field

jenniferarnesen:
I've linted these files in anticipation of making some changes. Also, renamed the text-editor-field component to "TextEditorField" instead of "ProgramRuleConditionField"


varl:
Looks good to me. 👍

PR #352, opened 2017-09-18
Improve the lintchanged script to cover untracked files

jenniferarnesen:


jenniferarnesen:
I don't claim to be a savvy shell script writer, so improvements welcome. Maybe it should have been written in javascript... 😸

nicolayr:
One challenge is that we probably want to use this exact script in multiple apps, right? So if we keep it in the app repo we'll have to copy it to every repo, which will be a maintenance nightmare.

Perhaps it would be a better option to keep this script in the `eslint-config-dhis2` package in stead?

jenniferarnesen:
I'll move it to eslint-config-dhis2, and add it as an executable in that npm package.

PR #355, opened 2017-09-20
Dhis2 2080 skip validation

Birkbjo:
Also contains organisationUnitLevels in validationRule.


src/forms/form-fields/multi-select.js
```diff
@@ -209,7 +213,9 @@ export default React.createClass({
     },
 
     _assignItems(items) {
-        if (this.props.referenceProperty === 'aggregationLevels') {
+        debugger;
```
nicolayr: ?

Markionium: @Birkbjo I knew we had a `debugger` statement... 

Also not sure why our eslint config lets that pass, but won't accept console.log? Thats probably something to look into. @jenniferarnesen

PR #365, opened 2017-10-17
Lint a set of files

jenniferarnesen:
For this review, I chose files that I know will be changed when the SQL view implementation is finished.

### DIFF
--- a/src/List/ContextActions.js
+++ b/src/List/ContextActions.js
@@ -1,7 +1,7 @@
 import { Subject } from 'rxjs';
 import log from 'loglevel';
 
-import { config, getInstance as getD2 } from 'd2/lib/d2';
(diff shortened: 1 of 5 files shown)


jenniferarnesen:
In List.component.js, around line 108, the propTypes are defined. But I'm not sure if all of them are actually required, or if I should make them optional, then add a defaultProps definition.

varl:
>In List.component.js, around line 108, the propTypes are defined. But I'm not sure if all of them are actually required, or if I should make them optional, then add a defaultProps definition.

If the same props are always passed in, default props would simplify the code for the callers and be a bit cleaner. If everyone just ends up passing in different stuff it doesn't matter that much.

jenniferarnesen:
Old and outdated.

PR #368, opened 2017-11-09
Include access when fetching the first org unit in a new DB

nicolayr:
Should fix the bug that makes it impossible to create the org unit tree in a fresh DB.


varl:
I resolved the conflict, looks good to me.

PR #379, opened 2018-01-05
Auth-mixin to hoc

Birkbjo:
Auth.mixin was removed from d2-ui. This PR implements Auth as a HOC instead.

Also fixes some issues related to removed index files from d2-ui.

### DIFF
--- a/src/EditModel/SharingNotification.component.js
+++ b/src/EditModel/SharingNotification.component.js
@@ -11,7 +11,7 @@ const SharingNotification = React.createClass({
         style: React.PropTypes.object,
     },
 
-    mixins: [Auth, Translate],
+    mixins: [Translate],
 
     render() {
         const createPublic = this.props.getCurrentUser().canCreatePublic(this.props.getModelDefinitionByName(this.props.modelType));
--- a/src/EditModel/SharingNotification.component.js
+++ b/src/EditModel/SharingNotification.component.js
@@ -1,9 +1,9 @@
 import React from 'react';
-import Auth from 'd2-ui/lib/auth/Auth.mixin';
 import Translate from 'd2-ui/lib/i18n/Translate.mixin';
 import { config } from 'd2/lib/d2';
 import FontIcon from 'material-ui/FontIcon/FontIcon';
 import Paper from 'material-ui/Paper/Paper';
+import { withAuth } from "../utils/Auth";
 
 const SharingNotification = React.createClass({
     propTypes: {
@@ -14,8 +14,8 @@ const SharingNotification = React.createClass({
     mixins: [Auth, Translate],
 
     render() {
-        const createPublic = this.getCurrentUser().canCreatePublic(this.getModelDefinitionByName(this.props.modelType));
-        const createPrivate = this.getCurrentUser().canCreatePrivate(this.getModelDefinitionByName(this.props.modelType));
+        const createPublic = this.props.getCurrentUser().canCreatePublic(this.props.getModelDefinitionByName(this.props.modelType));
+        const createPrivate = this.props.getCurrentUser().canCreatePrivate(this.props.getModelDefinitionByName(this.props.modelType));
         const notificationStyle = Object.assign({}, this.props.style, {
             background: 'none',
             margin: '14px 0 0 -4px',
@@ -45,4 +45,4 @@ const SharingNotification = React.createClass({
     },
 });
 
-export default SharingNotification;
+export default withAuth(SharingNotification);
--- a/src/EditModel/SharingNotification.component.js
+++ b/src/EditModel/SharingNotification.component.js
@@ -14,8 +14,9 @@ const SharingNotification = React.createClass({
     mixins: [Translate],
 
     render() {
-        const createPublic = this.props.getCurrentUser().canCreatePublic(this.props.getModelDefinitionByName(this.props.modelType));
-        const createPrivate = this.props.getCurrentUser().canCreatePrivate(this.props.getModelDefinitionByName(this.props.modelType));
+        const modelDef = this.props.getModelDefinitionByName(this.props.modelType);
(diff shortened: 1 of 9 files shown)


src/utils/Auth.js
```diff
@@ -0,0 +1,22 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { getInstance } from 'd2/lib/d2';
+
+
+export const withAuth = (Component) => {
+
+    const Auth = (props, { d2 }) => {
+        const extraProps = {
+            getCurrentUser: () => d2.currentUser,
```
Markionium: I think there wasn't really a reason why this was a function.  It might as well just be `currentUser`.

src/utils/Auth.js
```diff
@@ -0,0 +1,22 @@
...
+            getModelDefinitionByName: (modelType) => d2.models[modelType]
+
+        }
+        return <Component {...props} {...extraProps} />
+    }
+    Auth.contextTypes = {
+        d2: PropTypes.object
+    }
+    return Auth;
+}
```
Markionium: It might be nice to add a displayName based in the original component for dev niceness. 
https ://reactjs.org/docs/react-component.html#displayname

src/EditModel/SharingNotification.component.js
```diff
@@ -45,4 +45,4 @@ const SharingNotification = React.createClass({
     },
 });
 
-export default SharingNotification;
+export default withAuth(SharingNotification);
```
Markionium: A HoC is much nicer. :) I'm glad someone finally got around to doing that.

src/utils/Auth.js
Birkbjo: Only reason I kept it like this was to keep the "API" the same. As the old mixin used a function, but I agree.

Markionium: Yeah i thought as much, totally makes sense. My comment was more a general observation :P

src/EditModel/SharingNotification.component.js
```diff
@@ -1,21 +1,21 @@
...
     },
 
-    mixins: [Auth, Translate],
+    mixins: [Translate],
 
     render() {
-        const createPublic = this.getCurrentUser().canCreatePublic(this.getModelDefinitionByName(this.props.modelType));
-        const createPrivate = this.getCurrentUser().canCreatePrivate(this.getModelDefinitionByName(this.props.modelType));
+        const createPublic = this.props.getCurrentUser().canCreatePublic(this.props.getModelDefinitionByName(this.props.modelType));
+        const createPrivate = this.props.getCurrentUser().canCreatePrivate(this.props.getModelDefinitionByName(this.props.modelType));
```
varl: Please extract `this.props.getModelDefinitionByName(this.props.modelType)` to a variable to reduce the line length, reduce code duplication, and make it a bit easier to read.

src/EditModel/SharingNotification.component.js
varl: Very slick. Much easier to see where the added logic comes from.

src/List/List.component.js
```diff
@@ -34,18 +33,19 @@ import Dropdown from '../forms/form-fields/drop-down';
 import DropdownAsync from '../forms/form-fields/drop-down-async';
 import { getFilterFieldsForType } from '../config/maintenance-models';
 import './listValueRenderers';
+import { withAuth } from "../utils/Auth";
 
 // Filters out any actions `edit`, `clone` when the user can not update/edit this modelType
 function actionsThatRequireCreate(action) {
-    if ((action !== 'edit' && action !== 'clone') || this.getCurrentUser().canUpdate(this.getModelDefinitionByName(this.props.params.modelType))) {
+    if ((action !== 'edit' && action !== 'clone') || this.props.getCurrentUser().canUpdate(this.props.getModelDefinitionByName(this.props.params.modelType))) {
```
varl: Please extract the argument for `canUpdate` for readability.

src/List/List.component.js
```diff
@@ -34,18 +33,19 @@ import Dropdown from '../forms/form-fields/drop-down';
...
         return true;
     }
     return false;
 }
 
 // Filters out the `delete` when the user can not delete this modelType
 function actionsThatRequireDelete(action) {
-    if (action !== 'delete' || this.getCurrentUser().canDelete(this.getModelDefinitionByName(this.props.params.modelType))) {
+    if (action !== 'delete' || this.props.getCurrentUser().canDelete(this.props.getModelDefinitionByName(this.props.params.modelType))) {
```
varl: Please extract the argument for `canDelete` for readability.

src/List/ListActionBar.component.js
```diff
@@ -26,20 +24,22 @@ const ListActionBar = React.createClass({
             zIndex: 10,
         };
 
-        const modelDefinition = this.getModelDefinitionByName(this.props.modelType);
+        const modelDefinition = this.props.getModelDefinitionByName(
+            this.props.modelType
+        );
```
varl: ☝️ 👍

src/utils/getDisplayName.js
```diff
@@ -0,0 +1,9 @@
+export default function getDisplayName(Component) {
+    return (
+        Component.displayName ||
+        Component.name ||
+        (typeof Component === 'string' && Component.length > 0
```
varl: Are there instances where we pass a string into the `getDisplayName` function? Passing in a string here and have it return the same string is basically a `toString` function.

I would prefer we could avoid the `typeof` check and just do `displayName || name || 'Unknown'`.

Markionium: Recompose has a helper for this one and we already have this dependency. https://github.com/acdlite/recompose/blob/master/docs/API.md#getdisplayname

varl:
@Birkbjo Nice. Ship it. :shipit: 👍

PR #380, opened 2018-01-06
Add new fields for TrackedEntityType

Birkbjo:
Adds two new fields, as described here: [DHIS2-2607](https://jira.dhis2.org/browse/DHIS2-2607)


src/config/field-config/field-order.js
```diff
@@ -235,6 +235,9 @@ const fieldOrderByName = new Map([
     ['trackedEntityType', [
         'name',
         'description',
+        'minAttributesRequiredToSearch',
+        'maxTeiCountToReturn'
+
```
varl: Any reason for the blank line inside the array?

PR #381, opened 2018-01-09
DHIS2-2683 Remove callbacks from the `onEnter` handlers

varl:
Due to how HashHistory is implemented when it changes URL hashes
in v3 of react-router multiple events fires when we use
`HashHistory.push(url)`. This in turn causes the `onEnter` handler
in react-router to run twice.

The XHR request itself is cached so no extra data is transmitted from
the server but another consequence is that the components render twice
which has performance implications.

For more information see the [JIRA
issue](https://jira.dhis2.org/browse/DHIS2-2683).

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -14,10 +14,9 @@ import store from './store';
 import { resetActiveStep } from './EditModel/actions';
 import { loadEventProgram } from './EditModel/event-program/actions';
 import { loadProgramIndicator } from './EditModel/program-indicator/actions';
+import { noop } from 'lodash/fp';
 
 import onDemand from './on-demand';
-
-function noop () {}
 
 function initState({ params }) {
     initAppState({
--- a/src/router.js
+++ b/src/router.js
@@ -17,6 +17,8 @@ import { loadProgramIndicator } from './EditModel/program-indicator/actions';
 
 import onDemand from './on-demand';
 
+function noop () {}
+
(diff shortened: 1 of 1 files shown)


src/router.js
```diff
@@ -138,56 +141,48 @@ function createLoaderForSchema(schema, actionCreatorForLoadingObject, resetActiv
...
+        // changed by adding an organisation unit which would need to be
+        // reflected in the organisation unit tree
         initState({ params });
-        return callback();
     }
 
     initState({ params });
     return listActions.loadList(params.modelType)
         .take(1)
         .subscribe(
```
varl: I'm pretty new to RxJS so I'm not sure if it makes sense to subscribe here anymore to capture the error messages, but I added a `noop` function to allow our error handling to function as it did before.

Birkbjo: I believe it does, as the loadList, and similar functions are still async (but handled by observables) and can throw an error, so this makes sense to me!

src/router.js
```diff
@@ -17,6 +17,8 @@ import { loadProgramIndicator } from './EditModel/program-indicator/actions';
 
 import onDemand from './on-demand';
 
+function noop () {}
```
Birkbjo: Could probably get this from lodash?

varl: Aight, fixed in 5ecac0d.

PR #383, opened 2018-01-17
WIP: Feature/dhis2 11 tracker programs

Birkbjo:
Features Tracker Program configuration.


src/EditModel/event-program/epics.js
```diff
@@ -9,70 +9,135 @@ import {
...
-                    .then(dataElements => dataElements.toArray())
+        const availableData$ = d2$.flatMap(d2 =>
+            Observable.combineLatest(
+                Observable.fromPromise(
+                    d2.models.dataElements
+                        .filter()
+                        .on('domainType')
+                        .equals('TRACKER')
+                        .list({ paging: false })
+                        .then(dataElements => dataElements.toArray())
```
varl: Maybe pull out the promises into their own variables?

```
...
Observable.combineLatest(
    Observable.fromPromise(dataElements,
    Observable.fromPromise(trackedEntityAttributes)
),
...
```

varl: Note to self: pick up from here tomorrow.

src/EditModel/event-program/create-data-entry-form/ActionButton.component.js
```diff
@@ -0,0 +1,22 @@
...
+import FontIcon from 'material-ui/FontIcon';
+
+const ActionButton = ({ onClick, icon }) => {
+    const noTransition = {
+        transition: 'none',
+    };
+
+    const onIconClick = (e) => {
+        if (e) e.stopPropagation();
+        onClick();
```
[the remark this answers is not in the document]
Birkbjo: @kjesvale

src/EditModel/event-program/epics.js
```diff
@@ -83,132 +148,311 @@ function loadEventProgramMetadataByProgramId(programId) {
...
+                            'programStageToEditCopy',
+                            null,
+                            eventProgramStore.getState()
+                        )
+                    );
+                } catch (e) {
+                    console.log(e);
+                }
+            })
+        )
```
varl: This nested function from L413-440 could be moved into a named function and the logic itself could be split into smaller functions as there are multiple things happening. One function could look for the id, and pass it to the next which does the state operations.

I also find it a bit strange to first log out that the id doesn't exist on L420-422, and then use that non-existing id to set something. Wouldn't that operation fail and log out the error on L437?

src/EditModel/event-program/reducers.js
```diff
@@ -3,35 +3,59 @@ import {
...
+    state = { activeStep: first(steps), disabled: false, isLoading: true },
+    action
+) {
     switch (action.type) {
-    case EVENT_PROGRAM_STEP_CHANGE:
-        return {
-            activeStep: action.payload,
-        };
+        case EVENT_PROGRAM_STEP_CHANGE:
+          //  const disabled = action.payload == ""
```
varl: I see that this is commented out, but another trivia question.

If this was not commented, in what scope would the `const disabled` declared?
What are the effects on the surrounding code?
Are you able to reuse the `disabled` variable elsewhere?
Can you mitigate the issues?

(See this for the explanation: https://eslint.org/docs/rules/no-case-declarations)

src/EditModel/event-program/tracker-program/assign-tracked-entity-attributes/epics.js
```diff
@@ -0,0 +1,82 @@
...
+            const removeAttributes = keepProgramAttributesNotInUidList(attributeIdsToRemove);
+            program.programTrackedEntityAttributes = removeAttributes(programAttributes);
+            store.setState({
+                ...store.getState(),
+                program,
+            });
+        })
+        .mapTo({ type: PROGRAM_ATTRIBUTES_ADDREMOVE_COMPLETE });
+
+const isObjectHasId = id => compose(isEqual(id), get('id'));
```
varl: `isObjectHasId`? This doesn't make sense to me. 🤕

src/Snackbar/reducers.js
```diff
@@ -11,6 +11,7 @@ const initialSnackState = {
 
 export default function snackReducer(state = initialSnackState, action) {
     switch (action.type) {
+        case "NOTIFY_USER":
```
varl: Indendation is mismatched in this `switch`.

src/EditModel/event-program/tracker-program/assign-tracked-entity-attributes/epics.js
Birkbjo: Haha. This is not something I have come up with. Part of the refactor to move some files around.

varl: Could you rename it so the name makes sense? If it returns true/false it could be `isObjectWithId` so it at least is parses as English.

varl:
Reviewed set 7fd72c1 through f91f83e. LGTM.

src/EditModel/event-program/assign-data-elements/AssignDataElements.js
```diff
@@ -233,7 +233,7 @@ function AssignDataElements(props, { d2 }) {
 
     return (
         <Paper>
-            <div style={{ padding: '2rem 3rem 4rem' }}>
+            <div style={{ padding: '2rem 3rem 4rem', ...(props.outerDivStyle)}}>
```
Birkbjo: Conditional spread ftw

PR #384, opened 2018-01-24
Dataset notifications ui

paragateoslo:


src/config/field-overrides/programNotificationTemplate.js
```diff
@@ -56,6 +56,7 @@ export default new Map([
     }],
     ['recipientUserGroup', {
         component: (props) => {
+            console.log(props);
```
varl: 👆 Is this a development log line? Either remove it before merging to master or replace it with the log module used for logging elsewhere in the app.

paragateoslo: Ah good catch, that was just for temporary debugging

varl:
LGTM. 👍

Birkbjo:
Looks good. Need to fix conflict though.

PR #385, opened 2018-01-24
Bug/field rule for hide field not working with stepper

paragateoslo:
Fix for the stepper that would ignore HIDE_FIELD/SHOW_FIELD rules with config/field-config/field-groups.js.

Since the field rules in EditModel/form-rules/index.js uses the same way to hide and show
fields with the HIDE_FIELD and SHOW_FIELD rule, this causes the rule not work using 
the stepper. 

The EditModel used two ways to hide a field. One involved setting the style to 'display: none' to hide fields that receives style from props. The second was using the same as rules by saving the component into the "hiddenComponent" property and then deleting the content of "component".
      
Now rules uses the second method and EditModel->stepper using the display: none method. 

For fields to hide properly with the stepper, they therefore need to receive styles in props.

### DIFF
--- a/src/EditModel/EditModelForm.component.js
+++ b/src/EditModel/EditModelForm.component.js
@@ -225,7 +225,7 @@ export default React.createClass({
                 activeStep: step,
                 fieldConfigs: this.state.fieldConfigs.map((field) => {
                     if (stepsByField[field.name] === step) {
-                        if (field.hiddenComponent) {
+                        if (field.hiddenComponent && !field.hasHiddenRule) {
                             field.component = field.hiddenComponent;
                             field.hiddenComponent = undefined;
                         }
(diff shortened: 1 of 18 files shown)


src/EditModel/EditModelForm.component.js
```diff
@@ -218,14 +221,33 @@ export default React.createClass({
...
+        When the stepper is set to a step in which the field is supposed to be rendered
+        the hiddenComponent will be set back to the component variable.
+
+        **HACK**
+        Since the field rules in EditModel/form-rules/index.js uses the same way to hide and show
+        fields with the HIDE_FIELD and SHOW_FIELD rule, this causes the rule not work using 
+        the stepper. Setting the style to 'display: none' will would work for the stepper and 
+        still allowing the rules to use "hiddenComponent" as a way to show and hide fields, 
+        but since not all component does not spread and merge style, it will not always work.
+        Whe all components uses spread style this hack can be removed.
```
varl: Did you investigate how many components there are which need to use the spread style hack?

varl:
Nice, thanks for fixing the spread styles on the components to avoid hacking it.

varl:
LGTM. 👍

PR #387, opened 2018-01-26
Fix creating new programstage.

Birkbjo:
ModelCollectionProperty.remove wants a model object, not ID. Fixed this.
Some cleanup.


src/EditModel/event-program/tracker-program/epics.js
```diff
@@ -132,34 +157,31 @@ export const cancelProgramStageEdit = action$ =>
...
-                if (store.programStageToEditCopy.name === undefined) {
-                    const removedFromProgramStages = store.programStages.filter(
-                        (p, i) => i !== index
-                    );
-                    program.programStages.remove(stageId);
-                    programStageSetter = set(
-                        'programStages',
-                        removedFromProgramStages,
+                    if (index < 0) {
+                        console.warn(
```
varl: You use `console` to log in this file and `loglevel` in the next. I think `console` is fine for development, but we should probably use `loglevel` for logs we want to ship to clients.

PR #389, opened 2018-01-30
Adds a analyticsPeriodBoundaries component for program indicators

varl:
Related to https://jira.dhis2.org/browse/DHIS2-2496

_N.B. This depends on the extended datamodel targeted for 2.29, I will tag 2.28 in maintenance-app before merging this._

Some notes:

1. I could not figure out how to get the potential values for the enums from the API (I was told that it was not possible) so resorted to hardcoding them for `boundaryTarget` and `boundaryType`. I saw this technique being used elsewhere, e.g. `programRuleActionDialog.component.js`.

2. Updating subfield (add/remove/change value for nested dropdowns) need to be handled by shortcircuiting the `onChange` handler from the parent form builder to avoid the programIndicator model getting reset, effectively throwing away any changes, everytime a subfield was added/removed/updated. The code has to keep track of what fields change and only put them on the model when it's time to hit "save".

3. The form components assume that they are supposed to be laid out in a vertical layout, but the proposed design in DHIS-2496 flips this assumption and wants the elements laid out horisontally, so that's why all the components are wrapped in a HOC with some custom style which fixes margins and layout.

### DIFF
--- a/src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
+++ b/src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
@@ -1,6 +1,14 @@
+import React, { PropTypes } from 'react';
+import { equals, compose } from 'lodash/fp';
+import withState from 'recompose/withState';
+import getContext from 'recompose/getContext';
+import { getInstance } from 'd2/lib/d2';
+import lifecycle from 'recompose/lifecycle';
+
 import DropDown from '../../../forms/form-fields/drop-down';
 import TextField from '../../../forms/form-fields/text-field';
 import PeriodTypeDropDown from '../../../forms/form-fields/period-type-drop-down';
+
 
 // definitions mapped to java enum
 // https://github.com/dhis2/dhis2-core/blob/markus-program-indicators-cohorts/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/AnalyticsPeriodBoundaryType.java#L37-L40
@@ -15,7 +23,7 @@ const periodBoundaryTypes = [
     },
     {
         text: 'After start of reporting period',
-        value: 'AFTER_END_OF_REPORTING_PERIOD'
+        value: 'AFTER_START_OF_REPORTING_PERIOD'
     },
     {
         text: 'After end of reporting period',
@@ -23,15 +31,12 @@ const periodBoundaryTypes = [
     }
 ];
 
-const boundaryTarget = [
+const boundaryTargets = [
     {text: 'Incident date', value: 'INCIDENT_DATE'},
     {text: 'Event date', value: 'EVENT_DATE'},
     {text: 'Enrollment date', value: 'ENROLLMENT_DATE'}
 ];
 
-function onChange (e) {
-    console.log('onChange', e);
-}
 
 function AnalyticsPeriodBoundary (props) {
     console.dir(props)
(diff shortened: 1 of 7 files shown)


src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
```diff
@@ -0,0 +1,182 @@
...
+
+    return onChange({ target: {
+        value: list.concat([newVal])
+    }});
+}
+
+function AnalyticsPeriodBoundaryList ({ d2, model, onChange }) {
+    const getTranslation = d2.i18n.getTranslation.bind(d2.i18n);
+
+    const boundaries = model.analyticsPeriodBoundaries.map((props, i) => (
```
Birkbjo: This is undefined when creating a new ProgramIndicator. Should be added as default when creating the model in program-indicator/epics.js

src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
```diff
@@ -0,0 +1,182 @@
...
+// dhis2-core/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/AnalyticsPeriodBoundaryType.java#L37-L40
+// as they are not exposed over the API
+function AnalyticsPeriodBoundary (props) {
+    const style = {
+        marginRight: '14px'
+    }
+
+    const getTranslation = props.d2.i18n.getTranslation.bind(props.d2.i18n);
+
+    const DD = withStyle(DropDown, style);
```
Birkbjo: Not loving these names.

src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
```diff
@@ -0,0 +1,182 @@
...
+    return (
+        <div style={{
+                display: 'flex',
+                flexWrap: 'wrap',
+                justifyContent: '',
+                alignItems: 'center'
+        }}>
+            <DD
+                style={ {marginRight: '14px'} }
+                labelText={getTranslation('analytics_boundary_target')}
```
Birkbjo: If you pass translateOptions={true} to DropDown, it will translate the options for you.

paragateoslo:
Legit!!!

src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
varl: Will update with hungarian notation.. ;o

src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
varl: Thanks! A bit obtuse passing in a translation key as the "text" value. I'll update my code.

src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
varl: Damn. I completely missed this. I'll update ze code.

src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
varl: 7d61b3c

src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
varl: c83d942

src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
varl: 6cfb7cc7f1a8e55154687fb3c283464ff7c52898

varl:
Tested against a 2.28 backend. It gracefully falls back. Merging.

PR #391, opened 2018-01-31
Fix/no back arrow on dataset sections

paragateoslo:


src/List/HelpLink.component.js
```diff
@@ -18,8 +23,25 @@ function getDocsVersion({ major, minor, snapshot }) {
     return `${major}.${minor}`;
 }
 
+/**
```
varl: Nice, thanks for adding some docs!

The high-level description of what the function accomplishes is missing. Something like the one on L43-44 is great and tells me a lot of what I can use the function for.

src/List/HelpLink.component.js
```diff
@@ -18,8 +23,25 @@ function getDocsVersion({ major, minor, snapshot }) {
     return `${major}.${minor}`;
 }
 
+/** 
+ * @param {string} key The mapping it currently is checking with 
+ * @param {string} variablesToReplace The variables that are to be replaced with a help link 
+ *
+ * @returns {string} The partial path that is to be matched`
+ */
+function getReplacedPath(key, variablesToReplace) {
```
varl: This function threw me for a little loop. I see that you mainly moved it around but bear with me.

I sent in my key "foo" as the first param, `key`, and based on the documentation for `variablesToReplace` I should send a string as the second argument, so naturally I tried "bar". 

That didn't work so reading the code I could deduce that since you use `has` and `get` it is something like a Set or a Map and that let me explore a bit. 

My main assumption at this point was that if I send in a string, and a string of variables to replace, I will get a partially matched string match. I'm not sure what a partially matched string is at this point.

```
function getReplacedPath(key, variablesToReplace) {
    const placeholder = /\$\{(.+?)\}/g;
    return key.replace(placeholder, (match, variable) => {
        if (variablesToReplace.has(variable) &&
            variablesToReplace.get(variable)) {
            return variablesToReplace.get(variable);
        }
        return '.+?';
    });
}
// return: undefined

getReplacedPath('foobarmanchu', 'bar')
// return: "foobarmanchu"

getReplacedPath('foobar manchu', 'bar')
// return: "foobar manchu"

getReplacedPath('foo${bar}manchu', 'bar')
// return: TypeError: variablesToReplace.has is not a function

getReplacedPath('foo${bar}manchu', new Map([['bar', 'buz'],['foo', 'fuz']]))
// return: "foobuzmanchu"

getReplacedPath('foo${pants}manchu', new Map([['bar', 'buz'],['foo', 'fuz']]))
// return "foo.+?manchu"
```
1. So a couple of notes on the doc lines:

>The mapping it currently is checking with 
I'm not entirely sure what "mapping" or "it" is being referred to here since the variable is named "key". Would "key to replace" be an accurate interpretation?

Suggestion: "the template string with keys to replace, e.g. `foo${bar}`"

>The variables that are to be replaced with a help link 
Ok, so the first param is supposed to be a help link? Can it be any string or does it have to be a template string? How do I format my variables? Why is the second param a string? How do I send multiple values? In a comma separated list?

Suggestion: "map of values to interpolate"

>The partial path that is to be matched`

Why do I want the partial path which it is going to match? And if I am supposed to get a partial path, why do I get the whole modified string in return? Why if there is nothing to replace in my string would i want it replaced with `.+?`? If there's some reason for this specific behaviour please document that.

Suggestion: "the interpolated string"

2. `getReplacedPath` is not a declarative name, it does something _now_, it didn't do it in the past. 

Suggestion: `interpolate(str, values)`, `replace(str, values)`, something like that.

3. Given that my understanding is correct, the function can be written as:
```
function replace(str, variables) {
    return str.replace(/\$\{(.+?)\}/g, (_, v) => variables.has(v) ? variables.get(v) : '.+?');
}
```
Now, maybe max conciseness isn't the goal here, but we can at least reduce the amount of lookups it does; right now it first checks if the map `has` the var, then it sees if it can `get` the var, and if that is also truthy, it _throws away the var_ and `get`s it again.

PR #392, opened 2018-01-31
Add fields for event-programs

Birkbjo:
Add fields as described [here](https://jira.dhis2.org/browse/DHIS2-2608) for event-programs.


Birkbjo:
This is only relevant for Tracker Programs, so closing.

PR #398, opened 2018-02-05
Refactored List.component and DetailsBox

john-melin:
- Refactored List into a ES6 style component.
- Refactored DetailsBox into a ES6 style component.
- Refactored out DetailsBoxWithScroll from List to its own file.
- Moved all styles into one const.


Birkbjo:
Nice.

Kinda unrelated, but can you change the variable "this.props.params.modelType" on line 498 to "row.modelDefinition.name". For some reason it keeps the params from the old "render", and then fails to map the constants as it's the wrong modelType.

PR #401, opened 2018-02-09
Dhis2 2342/generate new uid for program rule actions

varl:


varl:
@Birkbjo I cannot find a way to separate the edit/clone operations when saving the changes to `programRule`. Do you know of one? I'm slightly concerned about orphans in the database since sending in a new `programRuleAction` UID to the metadata endpoint creates a new instance.

I talked to @Bekkalizer about this and he said that some of these relations actually clean up their own orphans, but `programRules` is not one of them.

varl:
Alright, I have verified that the orphaned `programRuleActions` are removed from the database when they are no longer referenced so this PR should be safe in that aspect.

varl:
Had an AFK discussion with @Birkbjo and while we agree that the solution for this category of issues should be solved server-side this is an OK solution in the interim. I added a comment to the code referencing the chain of issues related to cloning complex objects on frontend.

PR #412, opened 2018-02-21
Fix/rp229 3 mandatory tet

Birkbjo:
Fixes https://jira.dhis2.org/browse/RP229-3

Needed to restructure overrides to have custom rules for the different forms in the stepper.


varl:
LGTM :+1:

PR #414, opened 2018-02-21
Fix valueType being cleared when editing optionSet

Birkbjo:
This should fix https://jira.dhis2.org/browse/DHIS2-2782, and also the same issue for Attributes.


Birkbjo:
Note to self: should probably rename the function name, as we actually change the valueType based on the optionSet

PR #417, opened 2018-02-22
Fix/dhis2 2986 teta when changing tet

Birkbjo:
Should fix https://jira.dhis2.org/browse/DHIS2-2986.

For the stages to be applied, the user would need to select a type, and then go to the "Assign attributes"-step for the attributes to be added.

Attributes assigned this way are also not removed if the user goes back and changes the TET, as that would need a lot of logic to keep track of attributes added this way...

I still think this should happend back-end. But it works naively now.

### DIFF
--- a/src/EditModel/event-program/tracker-program/assign-tracked-entity-attributes/AssignAttributes.js
+++ b/src/EditModel/event-program/tracker-program/assign-tracked-entity-attributes/AssignAttributes.js
@@ -66,7 +66,6 @@ const mapDispatchToProps = dispatch =>
  * Extracts attributes that are in TrackedEntityTypeAttributes, but not in
  * programTrackedEntityAttributes
  * @param programModel Program model to use for programTrackedEntityAttributes
- * @param tetAttributes TrackedEntityType-attributes that we should check for.
  * @returns {Array} An array of TrackedEntityAttributes that are in tetAttributes, but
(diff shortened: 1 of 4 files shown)


src/EditModel/event-program/tracker-program/assign-tracked-entity-attributes/AssignAttributes.js
```diff
@@ -32,64 +42,105 @@ const styles = {
...
+ * not in programTrackedEntityAttributes, or empty if none.
+ */
+export function tetAttributesNotInProgram(programModel) {
+    if (
+        programModel &&
+        has('trackedEntityType.trackedEntityTypeAttributes', programModel)
+    ) {
+        return programModel.trackedEntityType.trackedEntityTypeAttributes.filter(
+            teta => {
+                console.log(teta);
```
varl: ?

src/EditModel/event-program/tracker-program/assign-tracked-entity-attributes/AssignAttributes.js
```diff
@@ -32,64 +42,105 @@ const styles = {
...
+                const hasAttribute = programModel.programTrackedEntityAttributes.find(
+                    ptea =>
+                        ptea.trackedEntityAttribute.id ==
```
varl: Any reason for the loose equality? If not, go with the strict.

PR #420, opened 2018-02-27
Fix where pattern would lose focus on slow input

john-melin:
- refactored trackedEntityAttribute to use field-rules rather than own custom in component "rules"
- refactored out the OrgUnitScope drop down into its own file.
- Added translation for Pattern
- linting of related files

### DIFF
--- a/src/forms/form-fields/check-box.js
+++ b/src/forms/form-fields/check-box.js
@@ -33,6 +33,7 @@ const CheckboxWrap = (props) => {
         isRequired,
         options,
         model,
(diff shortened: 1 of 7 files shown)


src/forms/form-fields/check-box.js
```diff
@@ -33,6 +33,7 @@ const CheckboxWrap = (props) => {
         isRequired,
         options,
         model,
+        undefined,
```
Birkbjo: ?

PR #422, opened 2018-02-27
Prevent removal of PTEAs defined in TETA

Birkbjo:
Prevents the ability to remove programTrackedEntityAttributes in "Assign attributes"-tab that are defined in the selected trackedEntityType.

The TETAs are loaded after the initial loading of program metadata, as we cannot know if it's an event- or tracker-program beforehand.

### DIFF
--- a/src/forms/form-fields/drop-down-async.js
+++ b/src/forms/form-fields/drop-down-async.js
@@ -143,7 +143,6 @@ class DropDownAsync extends Component {
                     });
                 }
             })
-            .then(() => this.onChange({ target: { value: this.props.value ? this.props.value.id : this.props.value} }))
     }
 
     render() {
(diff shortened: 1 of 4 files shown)


src/forms/form-fields/drop-down-async.js
```diff
@@ -142,7 +142,8 @@ class DropDownAsync extends Component {
                         )),
                     });
                 }
-            });
+            })
+            .then(() => this.onChange({ target: { value: this.props.value ? this.props.value.id : this.props.value} }))
```
varl: `this.props.value ? this.props.value.id : this.props.value`?

Why return the falsey `props.value`? I think it would be more clear to return something specific.

Birkbjo: To be honest, it's just because the way the value are passed down to the onchange handler to the DropDown. See line 185. 
I will change this either way though.

varl: Oh ok, it just looked a bit strange.

PR #428, opened 2018-03-04
Add missing fields to eventProgram.

Birkbjo:
Should fix https://jira.dhis2.org/browse/DHIS2-2457
Add translation for Snackbar through notifyUser()-action.
Also fixes a bug where saving without any changes did not reset saving-state.

We need two different form-builders, as they operate on two different object. The new fields are related to the "programStage".

### DIFF
--- a/src/EditModel/event-program/EventProgramStepperContent.js
+++ b/src/EditModel/event-program/EventProgramStepperContent.js
@@ -10,26 +10,46 @@ import AssignDataElements from './assign-data-elements/AssignDataElements';
 import EditDataEntryForm from './create-data-entry-form/CreateDataEntryForm.component';
 import AssignOrganisationUnits from './assign-organisation-units/AssignOrganisationUnits';
 import EventProgramNotifications from './notifications/EventProgramNotifications';
-import { createFormFor } from '../formHelpers';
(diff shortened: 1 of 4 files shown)


src/EditModel/event-program/EventProgramStepperContent.js
```diff
@@ -10,26 +10,46 @@ import AssignDataElements from './assign-data-elements/AssignDataElements';
...
 
-    const mapDispatchToProps = dispatch =>
-        bindActionCreators({ editFieldChanged }, dispatch);
+let ProgramStageForm = props => (
+    <FormBuilder
+        fields={props.fieldConfigs}
+        onUpdateField={props.editProgramStageFieldChange}>
+    </FormBuilder>
+)
+ProgramStageForm = createFieldConfigsFor('programStage', eventProgramStageFields, undefined, null, true, 'eventProgramStage')(ProgramStageForm);
```
varl: When there are so many parameters we should consider using a config object instead.

Birkbjo: I've thought about it, but I did not want to change it now as it's used all over the place.

PR #430, opened 2018-03-05
Wrap d2ui elements so that updates to d2ui does not break the UI

varl:


varl:
Move all the style to a wrapping `div` and set a base size for the elements.

PR #434, opened 2018-03-12
Fix custom form. Add custom registration form

Birkbjo:
Fixed a bunch of issues with custom forms by doing some refactoring.

There is still some refactoring that should be done, as EditCustomFormProgramStage and EditCustomRegistrationForm contains a lot of the same code, however we need the fix for 2.29.

DataEntryForm are used in dataSet, programsStages and program (custom registration form). 
It used the CKeditor, and I tried to extract common logic into a util-file. 
I did not touch the dataSet custom form, but it has a lot of the same code... However it would take me many hours to include that as well (but less now, as the util file should be somewhat generic).

Should fix these issues: 
https://jira.dhis2.org/browse/DHIS2-3167
https://jira.dhis2.org/browse/DHIS2-3199
https://jira.dhis2.org/browse/DHIS2-3225

### DIFF
--- a/src/EditModel/event-program/data-entry-form/EditCustomFormProgramStage.js
+++ b/src/EditModel/event-program/data-entry-form/EditCustomFormProgramStage.js
@@ -19,6 +19,8 @@ import eventProgramStore from '../eventProgramStore';
 import CKEditor from './CKEditor';
 import '../../../../scss/EditModel/EditDataEntryFormProgramStage.scss';
 import { getProgramStageDataElementsByStageId } from "../notifications/selectors";
+import PaletteSection from './PaletteSection';
+import { bindFuncsToKeys, moveEditorSelection, processFormData, insertElement as insElem, elementPatterns } from "./dataEntryFormUtils";
 
 const programStageDataElementWithProgramId = programStageId => programStageDataElement => ({
     id: `${programStageId}.${programStageDataElement.id}`,
@@ -27,6 +29,8 @@ const programStageDataElementWithProgramId = programStageId => programStageDataE
 
 const inputPattern = /<input.*?\/>/gi;
 const dataElementCategoryOptionIdPattern = /id="(\w*?)-(\w*?)-val"/;
+
+const PurePaletteSection = PaletteSection;
 
 const styles = {
     heading: {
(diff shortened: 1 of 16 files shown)


src/EditModel/event-program/data-entry-form/EditCustomFormProgramStage.js
```diff
@@ -146,20 +121,14 @@ class EditDataEntryForm extends React.Component {
...
-        this.disposables.add(Observable.combineLatest(
-            firstProgramStage$,
-            programStageDataElements$,
-            Observable.fromPromise(context.d2.Api.getApi().get('system/flags')),
-            (programStage, ops, flags) => ([programStage, ops, flags])
-        )
-            .take(1)
-            .subscribe(([programStage, programStageDataElements, flags]) => {
+        const { programStage, programStageDataElements } = props;
+            // Load flags
```
varl: This block (L125-174) is at the wrong level.

src/EditModel/event-program/data-entry-form/EditCustomFormProgramStage.js
```diff
@@ -146,20 +121,14 @@ class EditDataEntryForm extends React.Component {
...
+            this.disposables.add(Observable.fromPromise(context.d2.Api.getApi().get('system/flags'))
+            .subscribe(flags => {
```
varl: This function should be split up into smaller ones, right now it has many responsibilities (flags, operands, creating inserter functions, filtering elements, etc, that do not need to be inside the `next` callback.

I think only these two things in the function actually use the parameter:

```
[..].subscribe(flags => {
                 this.flags = flags.reduce((out, flag) => {
                     out[flag.path] = flag.name; // eslint-disable-line
                     return out;
                 }, {});

                 Object.keys(this.flags).forEach((flag) => {
                     insertFn[flag] = this.insertFlag.bind(this, flag);
                 });
});
```

The inserter functions could also be abstracted away to a function like `inserter(...)` which could be reused for operands and flags, making it a bit simpler as well.

I like functions which do one thing. I think they are a lot easier to reason about. This functions does many things. :)

varl:
Tested some more, seems good to me.

PR #436, opened 2018-04-06
DHIS2-1932 added section for standard variables in program rule expre…

john-melin:
…ssion
- Linting of related files.
- Added translation for standard variables.


varl:
LGTM.

PR #438, opened 2018-04-06
DHIS2-3251 fixed cramped dataSet group editor for dataSetSections

john-melin:
- Removed style from textField that was causing the bad positioning. 
- Refactored functions to use arrowfunctions rather than .bind();
- Linting in SectionDialog.component.js


varl:
@john-melin This does a lot more than just add some whitespace to the GUI. Please update the description with some more information of what it does.

PR #447, opened 2018-04-17
Program assignment in orgunit-edit

Birkbjo:
Adds the ability for assigning programs in the origanisation unit edit-view.

This is so that it's faster to assign multiple programs to an orgunit, instead of manually going through the programs and adding orgunits.

See https://jira.dhis2.org/browse/DHIS2-2356.

Need to go to System Settings -> Access -> Check "Allow assigning object to related objects during add or update" for the programs to show up in organisation units.

### DIFF
--- a/src/EditModel/objectActions.js
+++ b/src/EditModel/objectActions.js
@@ -155,12 +155,12 @@ objectActions.saveObject
         const d2 = await getInstance();
         const organisationUnit = modelToEditStore.getState();
 
-        if (!organisationUnit.isDirty() && !organisationUnit.dataSets.isDirty()) {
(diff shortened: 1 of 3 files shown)


src/EditModel/objectActions.js
```diff
@@ -155,12 +155,12 @@ objectActions.saveObject
         const d2 = await getInstance();
         const organisationUnit = modelToEditStore.getState();
 
-        if (!organisationUnit.isDirty() && !organisationUnit.dataSets.isDirty()) {
+        if (!organisationUnit.isDirty() && !organisationUnit.dataSets.isDirty() && !organisationUnit.programs.isDirty()) {
```
varl: Ok, but this is the last one we add before refactoring?

PR #448, opened 2018-04-17
Add date to send field to send message-action

Birkbjo:
See https://jira.dhis2.org/browse/DHIS2-3343


varl:
There is no fix version in JIRA, is it 29 or 30?

Birkbjo:
It's a subtask. Maintask is set for 2.30

PR #449, opened 2018-04-17
Program indicator clone

Birkbjo:
The bug was caused by AnalyticsPeriodBoundaries not generating new uids.

Implemented a "custom" handler function when cloning, as this will problably come up again in the future.

The UIDS are cleared when cloning, so the server will generate them. 

See
https://jira.dhis2.org/browse/DHIS2-3336


varl:
Looks much better. 👍

PR #452, opened 2018-04-21
DHIS2-3505 TET new property - allowAuditLog

abyot:


Birkbjo:
This is for 2.30 only right?

abyot:
Yes, that is for 2.30. I will merge the server side... then we can merge
this one.


---
Thank you,
Abyot.

abyot:
perfect... was this deployed? I don't see the effect of the merge.



---
Thank you,
Abyot.

PR #453, opened 2018-04-23
Fixes bug that would cause OrgUnittree in sidebar to always select fallback

HendrikThePendric:
- appState.state.selectedOrganisationUnit is a model, not an array
- because the condition included a check for .length the method loadSelectedOrganisationUnitState would never return the selectedOrganisationUnit but the users root instead

### DIFF
--- a/src/App/appStateStore.js
+++ b/src/App/appStateStore.js
@@ -99,7 +99,7 @@ async function getCurrentUserOrganisationUnits(disableCache = false) {
 }
 
 async function loadSelectedOrganisationUnitState() {
-    if (appState.state && appState.state.selectedOrganisationUnit && appState.state.selectedOrganisationUnit.length) {
+    if (appState.state && appState.state.selectedOrganisationUnit) {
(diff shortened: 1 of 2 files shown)


varl:
Can you make the tree in the sidebar to stay expanded when going back too?

dhis2-331

varl:
Looks good. Can you make the tree in the sidebar stay open (or re-open) when you go back?

dhis2-331

HendrikThePendric:
Good suggestion. Done.

varl:
Looks good to me!

@Birkbjo is it safe to merge in maintenance or is it still setup against 2.29?

Birkbjo:
Is this for 2.30 only? Would like to have the state-fix in 2.29 if it's backwards compatible

HendrikThePendric:
I only tested it in 2.29. Can't imagine any reason why it wouldn't work in 2.30. The solution is related to local state.

varl:
I updated the jira issue https://jira.dhis2.org/browse/DHIS2-331 to fix version 2.29.

PR #459, opened 2018-04-27
DHIS2-951 - Search across identifiers

Birkbjo:
See https://jira.dhis2.org/browse/DHIS2-951

A few problems I've not been able to solve, as we cannot combine "AND"- and "OR"-queries.

It works fine when doing just a search by the name, across name, shortName, id, code.
However, as many objects also have a filter for a type (like programType and aggregationType), and it's evaluated as an OR-query, the search will basically be disregarded as it matches all objects with the type.

I've a few ideas to fix this, but none of them are ideal.

We could do the filtering of the other filters than the search on the client (only when search is specified).
We could send multiple request to the server, and use the intersection of the results.

Morten mentioned having an own d2ql language for this, but that is way further down the line.


Birkbjo:
This is now updated to use the new 'identifiable' search-term.

PR #460, opened 2018-04-29
Fix/validation on save

john-melin:
DHIS2-3588 
Fix where user could not save any changes made to custom components in forms due to the "valid" prop in EditFormModel being set to false on default and not activating until all required fields where filled out. The reason for this not working is that many of the custom components do not notify the formBuilder onChange. And so the "valid" prop stay false and the save button remain disabled.

One would think that you would only need to add code to the onChange function of each components from formBuilder to fix it. But no. 

The formBuilder binds the onChange function with the field-property name. This causes issues when the name of the field-property is the same as one of the subComponents. See e.g 
SubjectMessageTemplateFields. This will cause all the other subComponents to also trigger changes on the field-property name at change.

So instead EditModelForms "valid" prop stays true and instead of relying on the formbuilder to post the form to the schema for validation, now validation is done when the user click Save. If the form is not valid, the snackBar will tell them and the save action is canceled.

This should work for all forms that uses EditModelForm. Validation does not work for programIndicators and programs as these create their own forms and doesn't have access to the formBuilder. 

- Added a function in EditField.saveAction to validate fields before saving and present the snackBar with what required field was invalid. 

- Added a setRequiredFieldsStepName in formHelpers that will add the name of the step a field is on. This will later be used to present the user with a helpful snackBar message on what step/group a required invalid field is on.
- Added the translatedLabelName to fieldConfigs. For error validation snackBar message
- Added functions in fields-groups to get a fields step belonging. 

- Added errorValidation texts to SubjectMessageTemplateFields and drop dows fields. 
- Refactored ProgramIndicator Components to their own files.
- Linting of related files.


Birkbjo:
Looks good to me, good job!
But it's a big PR and hard to backport. Should we just revert the default "valid" to true again for 2.28, 27 and 26? Or do you want to packport this to all those versions?

john-melin:
I think this shouldn't be that hard to backport without the refactoring and linting. I'll fix it during the day.

john-melin:
Backport PRs up. Btw 2.26 is not supported and was not back ported in the first place.

PR #461, opened 2018-04-29
DHIS2-3591 added the color value to the color prop

john-melin:


src/forms/form-fields/color-picker.js
```diff
@@ -29,7 +29,7 @@ export default function ColorPickerField(props) {
         <div>
             <div style={styles.field} >
                 <div style={styles.colorPickerLabel}> Color </div>
-                <ColorPicker color={'Select color'} {...props} onChange={transformChange} />
+                <ColorPicker color={props.value || 'Select color'} {...props} onChange={transformChange} />
```
varl: Can we just remove the "select color" string? Otherwise it needs to be i18n.

PR #468, opened 2018-05-09
Add preloadstate loader to routes

Birkbjo:
This fixes a number of issues.

When switching to dynamic loading with react-loadable, there was some issues with race conditions when changing routes, and the loading status in the list-component.
This is fixed by removing async-routes, as that fired the onLoad two times for each route change, being the origin of the race condition.

Some components use modelToEditStore.state directly, and may do so without checking if its initialised in the constructor. Therefore, I've added a HOC that checks for this together with loading the components. 

This should fix: https://jira.dhis2.org/browse/DHIS2-2683

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -18,6 +18,8 @@ import { loadEventProgram } from './EditModel/event-program/actions';
 import { loadProgramIndicator } from './EditModel/program-indicator/actions';
 import LoadableComponent, { LoadableWithLoaders } from './utils/LoadableComponent';
 
+listStore.subscribe(val => console.log(val));
+
 function initState({ params }) {
     initAppState({
         sideBar: {
@@ -148,7 +150,7 @@ function createLoaderForSchema(schema, actionCreatorForLoadingObject, resetActiv
     };
 }
 
-function loadList({ params, router: { replace } }) {
+function loadList({ params }, replace) {
     if (params.modelType === 'organisationUnit') {
         // Don't load organisation units as they get loaded through the
         // appState Also load the initialState without cache so we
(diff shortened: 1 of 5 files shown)


src/router.js
```diff
@@ -90,43 +95,41 @@ function loadObject({ params }, replace, callback) {
                 }, {});
 
             modelToEditStore.setState(Object.assign(modelToEdit, listFilters));
-            return callback();
         });
     } else {
         objectActions.getObjectOfTypeById({ objectType: params.modelType, objectId: params.modelId })
             .subscribe(
-                () => callback(),
+                () => {},
```
varl: Replace with `noop`.

src/router.js
```diff
@@ -260,14 +253,14 @@ const routes = (
...
-                    component={LoadableComponent({loader: () => import('./EditModel/EditDataSetSections.component')})}
+                    component={LoadableWithPreloadedStore({loader: () => import('./EditModel/EditDataSetSections.component')})}
                     onEnter={loadObject}
                     hideSidebar
                     disableTabs
                 />
                 <Route
                     path=":modelType/:modelId/dataEntryForm"
-                    component={LoadableComponent({loader: () => import('./EditModel/EditDataEntryForm.component')})}
+                    component={LoadableWithPreloadedStore({loader: () => import('./EditModel/EditDataEntryForm.component')})}
```
varl: To make it a bit more generic we could pass in the preloaded store to the component in addition to the loader function here, and use that in on L#37 in `LoadableComponent.js`.

Birkbjo: Can make it optional, and use the modelToEditStore default. Will be a bit verbose otherwise, imo.

varl: Yeah, but now the name should be: `LoadableWithPreloadedModelToEditStore` which is also a bit verbose. Doesn't matter a great deal so merge when you are ready.

Birkbjo:
I've now fixed some loading issues. The HOC is rewritten and simpler, and also ensures that the provided loading promises are resolved before loading the component, which caused a lot of issues.
It's also even more generic.

PR #481, opened 2018-05-31
Dhis2 3202/sql views

HendrikThePendric:
Pretty much all of this code applies only to the SQL Views functionality. Only exception is the the changes `src/EditModel/FormHeading.js`. This is an addition of a property so I could just feed it some text without the component trying to translate it.

When testing, please be aware that not all SQL queries can be executed/refreshed/viewed locally. I have had the most success with the *Code list data elements*. On my local machine, I could Execute and View this one.

For some reason, I can't seem to refresh any of my SQL views locally anymore. Yesterday I could refresh one or two. This is not to do with the maintenance-app, but with the queries themselves. You can see in the network tab that I am calling the right endpoint and getting a 409 back. So I would say the behaviour you will be seeing, i.e. the error snackbar message instead of the success one is expected behaviour.

I'll make sure I test again once this goes to play/dev.


src/SqlView/SqlView.js
```diff
@@ -0,0 +1,128 @@
...
+
+class SqlView extends Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            listGrid: null,
+        };
+    }
+
+    async componentWillMount() {
```
Birkbjo: Async stuff should probably be in componentDidMount. It's also deprecated, so we should try not to use it.

src/router.js
```diff
@@ -296,6 +296,12 @@ const routes = (
                 component={LoadableComponent({ loader: () => import('./OrganisationUnitHierarchy') })}
                 onEnter={initStateOuHierarchy}
             />
+            <Route
+                path="sqlViews/:modelId"
+                component={LoadableComponent({ loader: () => import('./SqlView') })}
```
Birkbjo: Can also use the LoadableWithLoaders, if you want to load something before the component is mounted.

src/SqlView/SqlView.js
HendrikThePendric: Done, see [this commit](https://github.com/dhis2/maintenance-app/pull/481/commits/5d71d2f4fe70679387c3160f1249c21f5a2f0592)

PR #488, opened 2018-06-07
Dhis2 980/locale ui management

HendrikThePendric:
All suggestions are welcome but the idea behind the current solution is as follows:
- Locales are not available in the schema endpoint so will not become available as a d2 model by default
- To make Locales behave consistently, we manually construct a custom model (as you will see, in practice it is actually a custom ModelDefinition).

To make this work I have: 
- Created a mechanism for attaching custom models to the existing d2 models when the app loads (`src/maintenance.js`, `src/config/custom-models/index.js`, `src/config/custom-models/locale/index.js`)
- Create a custom Schema and ModelDefinition for Locales (`src/config/custom-models/locale/localeSchemaDefinition.js`, `src/config/custom-models/locale/LocaleModelDefinition.js`)
- This was almost enough, but I had to also introduce two small fixes in stores / components (`src/EditModel/SingleModelStore.js`, `src/List/List.component.js` L515-524) 

The rest of the code you see are changes that would also be required for a normal d2-model.

Two other points of interest:
- The field-overrides part is ... uuuhhmm... interesting.... The problem was that the fields you want to show in a create-locale-form do not match the properties you want to write to. Because I have a custom `save` method in the ModelDefinition and had to use field-overrides anyway, this wasn't a problem (see `src/config/field-overrides/locale/*`, and the `save` method in `src/config/custom-models/locale/LocaleModelDefinition.js`)
- Because the locales endpoint doesn't support paging or filtering, I am caching the locales and have implemented client side filtering `src/config/custom-models/locale/LocaleModelDefinition.js`.


varl:
cc: @larshelge I want your thoughts on my comment.

varl:
cc: @Birkbjo go over my comment as well.

varl:
An interesting thing is that we could leverage this and migrate all `d2.models` to custom models in maintenance app and remove the tight coupling between DHIS2 and maintenance app, effectively giving us a path to refactor maintenance app into not being schema-driven-from-the-server without rewriting it.

varl:
Thinking and chatting some more with @HendrikThePendric about this has eased some of my concerns, but there is some additional cognitive load that comes with this. If it does work out well however, the upside is that we have a new mechanism which might prove powerful enough to do other improvements like decoupling from the backend.

LGTM :+1: :shipit: :rocket:

HendrikThePendric:
@varl Thanks for reviewing my code thoroughly. I do agree that this solution is far from ideal, but when considering the alternative, I considered this to be the lesser of the two evils. I will elaborate on this:

- In the maintenance app, pretty much every component/store relies on how d2 models work.
- So to put something in that isn't a d2 model, into the app something has got to change:

1. Either make adjustments sprinkled through the entire code base so that components/stores can deal with d2-models and other resources too.
2. Or make the other resources behave like d2-models.

I choose option 2 because out of the two, this one would have the lowest chance of regressions, as the vast majority of the new code is isolated.

Alternative ideas from @larshelge or @Birkbjo are very welcome :-)

PR #491, opened 2018-06-08
Ex: Move to sharing dialog in @dhis2/d2-ui

varl:
This is just to collaborate with @john-melin on a branch.

### DIFF
--- a/src/EditModel/event-program/tracker-program/program-stages/ProgramStageList.js
+++ b/src/EditModel/event-program/tracker-program/program-stages/ProgramStageList.js
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 
 import DataTable from 'd2-ui/lib/data-table/DataTable.component';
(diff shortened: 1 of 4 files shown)


src/List/List.component.js
```diff
@@ -578,6 +578,7 @@ class List extends Component {
                     </div>)}
                 {!!this.state.sharing.model &&
                     <SharingDialog
+                        d2={this.context.d2}
```
varl: Here it works fine.

src/EditModel/event-program/tracker-program/program-stages/ProgramStageList.js
```diff
@@ -153,6 +153,8 @@ class ProgramStageList extends Component {
         }
     }
 
+    // here you need to get d2 and pass it as a `d2` prop to sharing
+    // dialog, to do that you need to get d2 in here
```
varl: Here I ran into the error you mentioned where it just "hangs". I think that is because of RxJS swallowing the error, but most likely I think the error has to do with an incorrect prop is sent to the SharingDialog.

PR #492, opened 2018-06-10
Chore/program indicator refactor

john-melin:
# Part of the DHIS2-3231 validation issue

- Updating jest and babel to support dynamic imports with react loadable.
- Added eslint rules to not complain about lodash/fp import and nested parameter assign.
- Moved dependencies 

- Removed imports, functions and variables not in use
- Linting
- Separated fieldConfig functions into three separate files. The fieldConfigHelper was way too big (schemaFormCreator, schemaFieldConfigCreator and modelFieldConfigCreator)
- Refactored selectors to start with get according to redux convention

## The validation with forms using redux and redux observables had several issues
**Steps**
- When using steps the normal (non redux) EditModelForm passes a display:none style to components that are not part of the current step. This way you only need to handle 1 formBuilder. 

- The redux-observable forms using 1 formbuilder for each step. This makes validation handling difficult since you cant validate all fields for all steps at the same time, without having a reference to each formbuilder and looping through them (way harder than it needs to be).

The steps are therefore refactored to use the same method as non redux-observable forms by defining steps in a shared file in config/field-config/field-groups.js. The name of the steps will then be added to fieldConfigs to give the user useful validation in what step invalid fields reside. The hiding of fields are done with "getActiveStepFields()" in the program-indicator reducers.

- The functions concerning steps were moved to a separate step file. "EditModel/stepper/stepper.js"
- Moved arrowStepper to its own file
- Moved step actions from programIndicator to stepper.actions as they are not exclusive to programIndicators.

**SnackBar**
This required the FieldConfigs to be put in the redux store so they can "hidden" when changing steps. The tests later revealed that this was causing a circular dependency. This due to the fact that components stored in the fieldConfigs imports the redux store to get hold of the snackBar state, also residing in the redux store. 

I therefore had to refactor the SnackBar back to the old solution of storing the SnackBar state in its own "mark store (redux observable store)". 

**Validation**
To validate a field, you need a reference to a fieldConfig and the FormBuilder. (another reason for keeping the fieldConfigs in redux store). Previously we had neither. I therefore moved the formbuilder to the EditProgramIndicator class passing a reference of it to the saveAndValidate epic via an action.


varl:
I think this looks pretty good. Tested validation in program indicators and I didn't see any surprises. :+1:

PR #499, opened 2018-06-13
Add option group and optionsgroup set editmodels

Birkbjo:
WIP.

Works fine, but there are 15k+ options in the demo database. Need some way to lazy load the options in the multi-select.


varl:
Did a review again. Merge time?

PR #504, opened 2018-06-19
Open downloads in new window/tab

HendrikThePendric:
If it opens in a new tab is a browser setting. Could also open in a new window. This behaviour cannot be controlled via JS


varl:
At least has a higher chance of opening in a new tab if the user wants it to :^)

LGTM 👍

PR #506, opened 2018-06-20
Removed onEnter property  from dataEntryForm route

HendrikThePendric:
We are calling loadObject in the component property already with the correct footprint.


varl:
LGTM 👍

PR #508, opened 2018-06-21
DHIS2-2151-Sharing-UI

kjesvale:
Fixes [DHIS2-2151](https://jira.dhis2.org/browse/DHIS2-2151).

This PR introduces the new Access UI for tracker- and event programs, which lets the user change access/sharing settings for all program stages in one action.

Changes:
- Create folder "src/EditModel/event-program/program-access" and move the old "AssignOrganisationUnits" file to the more general "ProgramAccess.js"
- Create component "ProgramStagesAccess" and add the new UI in this file. This uses a new version of the sharing dialog that doesn't involve any networking requests. Instead, the "model" and store takes care of saving the new sharing/access settings.

Minor changes:
- Fetch "user:name" in program and program stages.
- Add new translations.
- src/EditModel/event-program/epics.js: Fix a bug where only the first program stage would be stored in a tracker program


kjesvale:
Will add translations before merging.

PR #510, opened 2018-06-25
Dhis2 3261/color palette in maintenance app

HendrikThePendric:


src/EditModel/option-set/OptionDialogForOptions/OptionDialogForOptions.component.js
```diff
@@ -78,6 +78,12 @@ const optionFormData$ = Observable.combineLatest(
     }))
     .flatMap(async ({ fieldConfigs, model, isAdd, ...other }) => {
         const d2 = await getInstance();
+
+        // Add modelId to fieldConfig so it can be accessed in the validator
+        fieldConfigs.forEach((fieldConfig) => {
```
john-melin: The model present in the setupFieldConfigs function above. This should rather be done there instead of setting up field configs in two places.

Also branch out into a function so comments are not needed.

HendrikThePendric: aha yes, good suggestion. I wanted to add it from there when I started, but was looking at the `modelToEdit` instead of the `model` and wrongly concluded that I didn’t have the option model in scope there. This last commit makes it all a lot more readable and logical. THANKS

PR #512, opened 2018-06-26
DHIS2-4128 Add an option to use 'custom' values, and if so, use a fre…

varl:
…e-text field

https://jira.dhis2.org/browse/DHIS2-4128

2018-06-26-170635_1749x692_scrot

### DIFF
--- a/src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
+++ b/src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
@@ -37,6 +37,9 @@ const boundaryTargets = [
     }
 ];
 
+// boundary type definitions mapped to java enum
+// dhis2-core/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/AnalyticsPeriodBoundaryType.java#L37-L40
+// as they are not exposed over the API
 const boundaryTypes = [
     {
         text: 'report_period_before_start',
@@ -60,8 +63,7 @@ function AnalyticsPeriodBoundary(props) {
     const getTranslation = props.d2.i18n.getTranslation.bind(props.d2.i18n);
 
     const filteredBoundaryTargets = boundaryTargets
-        .filter(e => e.value !== CUSTOM)
-        .filter(e => e.value === props.boundaryTarget)
+        .filter(e => e.value !== CUSTOM && e.value === props.boundaryTarget)
         .length
 
     const renderCustom = () => props.boundaryTarget !== null && filteredBoundaryTargets === 0;
--- a/src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
+++ b/src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
@@ -15,6 +15,7 @@ import {
     INCIDENT_DATE,
     EVENT_DATE,
     ENROLLMENT_DATE,
+    CUSTOM,
 } from './enums';
 
 const boundaryTargets = [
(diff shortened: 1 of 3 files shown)


src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
```diff
@@ -51,29 +56,52 @@ const boundaryTypes = [
...
-// as they are not exposed over the API
 function AnalyticsPeriodBoundary(props) {
     const getTranslation = props.d2.i18n.getTranslation.bind(props.d2.i18n);
+
+    const filteredBoundaryTargets = boundaryTargets
+        .filter(e => e.value !== CUSTOM)
+        .filter(e => e.value === props.boundaryTarget)
+        .length
+
+    const renderCustom = () => props.boundaryTarget !== null && filteredBoundaryTargets === 0;
```
varl: This covers the "no value" (`null`) option, and if the filtered list is empty, cases which we use to know if we need to render the free-text field.

src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
```diff
@@ -51,29 +56,52 @@ const boundaryTypes = [
...
-// boundary type definitions mapped to java enum
-// dhis2-core/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/AnalyticsPeriodBoundaryType.java#L37-L40
-// as they are not exposed over the API
 function AnalyticsPeriodBoundary(props) {
     const getTranslation = props.d2.i18n.getTranslation.bind(props.d2.i18n);
+
+    const filteredBoundaryTargets = boundaryTargets
+        .filter(e => e.value !== CUSTOM)
+        .filter(e => e.value === props.boundaryTarget)
+        .length
```
varl: The `boundaryTarget` can be:

- one of three enum values
- free-text
- special type of `CUSTOM` which only exists in this context

This creates a filtered list without the `CUSTOM` option and with the overlap between the stored `boundaryTarget` and the enums we have locally. With this we can proceed to figure out what to display.

HendrikThePendric:
Code looks good to me. Two things to perhaps consider:
1. The comments you placed above (especially the last one) are incredibly useful. I would consider leaving them in the code instead of in this PR. If we need to come back to this they will be helpful.
2. When I ran this branch locally, I got a PropTypes error. Probably not related to your work. But it is happening in the same view.

```
Warning: Unknown prop `status` on <span> tag. Remove this prop from the element. 
    in span (created by branch(branch(nest(span, mapProps(ActionCheckCircle)))))
    in branch(branch(nest(span, mapProps(ActionCheckCircle)))) (created by ProgramIndicatorExpression)
    in div (created by ProgramIndicatorExpression)
    .... (rest of stack trace omitted)
```

src/config/field-overrides/program-indicator/AnalyticsPeriodBoundaries.js
HendrikThePendric: Why not filter in one go? You are iterating through the array twice now. (even thought the array will be shorter after the first filter...
```
.filter(e => e.value !== CUSTOM && e.value === props.boundaryTarget)
```

varl: I did it out of readability concerns. I didn't like how the condensed predicate looked.

varl:
>The comments you placed above (especially the last one) are incredibly useful. I would consider leaving them in the code instead of in this PR. If we need to come back to this they will be helpful.

Thanks, I mistakenly removed it because I thought it said boundary *target* and not boundary *type*.

>When I ran this branch locally, I got a PropTypes error. Probably not related to your work. But it is happening in the same view.

Yeah, it's from `program-indicator/ExpressionStatusIcon.js`, it uses `mapProps` to nest components so the `status` prop winds up directly on a `span` tag, which isn't a valid HTML prop, hence the warning. I'm ignoring it because that code is a ratnest...

PR #513, opened 2018-06-27
Show parent when searching for leaf node

varl:
https://jira.dhis2.org/browse/DHIS2-2160

1. To avoid the problem where a user becomes "trapped" in a leaf node we can always show the parent in the search results
2. To avoid the problem where a user selects a leaf node and winds up with an empty list to the right, we can show the selected org unit in the list

The selected org unit is always shown in the right hand list, always at the top instead of at the bottom. Since the `modelCollection` is Map-backed some trickery is required (if you have a better idea let me know).

I patched d2-ui to retrieve the `access` field for children in the list. It's a bit annoying that some data is fetched through the component and some through the app. The app should ideally feed the component with data but this is the reality of the old d2-ui. New version is pushed to d2-ui@29.0.21.

Since I am using a home-grown root-node for the search result list I also need to add the fields required for the component to `treeRoot`. This is the minimum I could get away with since the component makes a lot of assumptions about the data passed in.

### DIFF
--- a/src/App/appStateStore.js
+++ b/src/App/appStateStore.js
@@ -69,7 +69,10 @@ async function getCurrentUserOrganisationUnits(disableCache = false) {
     }
 
     const d2 = await getInstance();
-    const organisationUnitsCollection = await d2.currentUser.getOrganisationUnits({ paging: false });
+    const organisationUnitsCollection = await d2.currentUser.getOrganisationUnits({
+        fields: ':all,access,displayName,path,children[id,displayName,path,children::isNotEmpty,access]',
+        paging: false
+    });
 
     if (d2.currentUser.authorities.has('ALL') && !organisationUnitsCollection.size) {
         const rootLevelOrgUnits = await d2.models.organisationUnits.list({
@@ -77,7 +80,7 @@ async function getCurrentUserOrganisationUnits(disableCache = false) {
             paging: false,
             fields: [
                 'id,displayName,path,publicAccess,access,lastUpdated',
-                'children[id,displayName,path,children::isNotEmpty]',
+                'children[id,displayName,publicAccess,access,path,children::isNotEmpty]',
             ].join(','),
         });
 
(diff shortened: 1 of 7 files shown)


src/App/appStateStore.js
```diff
@@ -69,15 +69,18 @@ async function getCurrentUserOrganisationUnits(disableCache = false) {
     }
 
     const d2 = await getInstance();
-    const organisationUnitsCollection = await d2.currentUser.getOrganisationUnits({ paging: false });
+    const organisationUnitsCollection = await d2.currentUser.getOrganisationUnits({
+        fields: ':all,access,displayName,path,children[id,displayName,path,children::isNotEmpty,access]',
```
HendrikThePendric: Are you sure you need `:all` fields?

src/List/organisation-unit-list/OrganisationUnitList.component.js
```diff
@@ -35,13 +35,14 @@ export default class OrganisationUnitList extends React.Component {
...
-                    // of available organisation units to pick from
-                    if (userOrganisationUnitIds.indexOf(selectedOrganisationUnit.id) >= 0) {
-                        organisationUnitList.add(selectedOrganisationUnit);
-                    }
+                    // DHIS2-2160 Add the selected node to the list to
+                    // avoid having to select the parent node to edit
+                    // the selected node...
+                    let filteredOrgUnits = organisationUnitList.toArray();
+                    filteredOrgUnits.unshift(selectedOrganisationUnit);
+                    let prependedOrgUnitList = ModelCollection.create(d2.models.organisationUnit, filteredOrgUnits);
```
HendrikThePendric: I have looked for an alternative way to do this, but also couldn't find anything better. There is the `valuesContainerMap` which has an `[[Entries]]` symbol. This is an array. God knows what happens when you unshift straight into that. The solution you have is correct and reads well. Performance could well be quite poor because you must be iterating through stuff at least twice, but I don't think this is an issue, since we are already working with a response that was filtered on the server.
I say go for it :-)

src/App/appStateStore.js
varl: Yes, it is absolutely necessary for the app to function.

I needed to override the fields passed to `getOrganisationUnits`, and only way to do that was to duplicate the `fields` found here: https://github.com/dhis2/d2/blob/master/src/current-user/CurrentUser.js#L212 so I could add `access` to that list (not sure if i should have expected to get it using `:all`, but I don't.

So this is actually done *every time* any app does a call to `d2.currentUser.getOrganisationUnits()`  without overriding the `fields` property. It's just more explicit now.

PR #521, opened 2018-07-02
DHIS2-4145-relationship type

Birkbjo:


src/config/field-overrides/relationshipType.js
```diff
@@ -0,0 +1,141 @@
...
+    };
+
+    render = () => {
+        return (
+            <div>
+                <SelectField
+                    fullWidth
+                    value={this.state.relationshipEntity}
+                    onChange={this.selectRelationshipEntity}
+                    floatingLabelText={this.translate('from_constraint')}
```
HendrikThePendric: This needs to by dynamic so it can display either "From constrain" or "To constraint"

Birkbjo: Thanks for noticing!

PR #522, opened 2018-07-03
Tech 93/tweak eslint configuration

HendrikThePendric:
**A summary of my work:**
- Switched from the `dhis2` ESLint config to a combination of `react-app` and `prettier` (the latter is added so that ESLint and Prettier don't disagree).
- Turned off three rules for various reasons. I'll elaborate below.
- Added a `.prettierrc` with a sensible configuration
- Let ESLint auto-fix fixable issues, which was all of the line-order related stuff
- I then went in and fixed all the straightforward issues in bulk, fixing mainly `no-unused-variables` and `eqeqeq` errors
- The more complex errors I fixed, tested, and committed one-by-one

**Disabled rules**
- `no-param-reassign`: I suppose disabling this rule is slightly ambiguous. My main argument for disabling is that it is just sometimes what needs to happen, for example [see here](https://github.com/dhis2/maintenance-app/blob/TECH-93/tweak-eslint-configuration/src/utils/LoadableComponent.js#L61-L63). There are only three cases of this, so if we want to enable this rule, we wouldn't have to fix a lot.
- `no-fallthrough`: I switched this one off because this is what happens in reducers quite often. Several different actions can lead to the same state transformation, i.e. `return: payload` where payload could be completely different things for different actions.
- `array-callback-return`: I'm not that happy turning off this rule, but it was giving lots of false positives. This rule expects the `.map` function to be the native `Array.prototype.map` and in that context it is a very good idea to expect the callback to always return something. However, we are also dealing with `rx/js` Observables, which also have a `.map` operator, which apparently doesn't need to return anything. If we switch this rule back on, we will have to fix a lot of epics. I am assuming the current state of the code is perfectly valid, so in the end I think disabling the rule is the only way to go....

**Prettier**
- I did include `prettier` as a dependency and have configured it, and ESLint, so that the two don't collide.
- I did not include a script in `package.json` that would prettify all of our code, because (as Viktor said) this would lead to too many file changes.

So the end result is that you can use Prettier to format your code on-demand. For example in VS-Code:
- Install the VS Code Prettier extention `ext install prettier-vscode`
- Make sure that you are not doing any auto-formatting on save or paste, i.e. make sure the following rules are applied:
`"editor.formatOnSave": false,`
`"editor.formatOnPaste": false,`
If you don't apply these rules, every file you work on will contain a lot of formatting related changes
- Now you can format on demand:
On a file: `Shift+Alt+F`
On a selected section: `Cmd+K,Cmd+F`


.eslintrc
```diff
@@ -1,11 +1,21 @@
...
+            "version": "^15.3.1"
+        }
+    },
     "env": {
         "browser": true,
         "jest": true
     },
     "rules": {
-        "no-param-reassign": [2, { "props": false }],
+        "no-param-reassign": "off",
```
john-melin: I think it might be unwise to have this completely off. As JS is pass by value, this rule catches the cases where parameters are directly being reassigned instead of nested properties. 

Only reassignment of nested properties should allowed.

HendrikThePendric: I tried switching the rule back on as we had it. This resulted in three errors. In all these cases, I would say that parameter re-assignment was a valid approach to solve the problem in question. Please review these and see what you think:
- https://github.com/dhis2/maintenance-app/blob/TECH-93/tweak-eslint-configuration/src/utils/LoadableComponent.js#L61-L63
- https://github.com/dhis2/maintenance-app/blob/TECH-93/tweak-eslint-configuration/src/EditModel/SingleModelStore.js#L150
- https://github.com/dhis2/maintenance-app/blob/TECH-93/tweak-eslint-configuration/src/EditModel/formHelpers.js#L195

I kind of agree with you that parameter re-assignment in general is a pretty bad idea. Perhaps what we could do, is just keep the rule as it was, and place ESLint disabling comments around the three parts of code listed above.

Thanks for the feedback. Looking forward to getting more of it 🔢

john-melin: Yeah. For the two first cases i do agree that the reassignment serves a purpose, at least in the way the code is constructed. The reassignment in the third case seems redundant as it's already handled in the parameters. 

I think parameter reassignment may often be an indicator that the code structure might be a bit off, and having the waring on in those cases requires the coder to consider the solution one more time. When adding ESlint disabling at least you have to make a conscious decision.

PR #523, opened 2018-07-04
DHIS2-3403 program notification parent

Birkbjo:
Adds checkboxes in program notification UI, when recipient is "User group"
See https://jira.dhis2.org/browse/DHIS2-3403


HendrikThePendric:
@Birkbjo What do you think of my solution: https://github.com/dhis2/maintenance-app/pull/523/commits/6fc3fcd0f1bbc852e958f9d6bdc522bbe1bad1f3

This solution meets all requirements and utilised existing mechanism in the maintenance-app.

On the other hand: the UI is a bit special.... I guess it really should have been implemented as an enum in the backend with the following values:
- `notifyAll`
- `notifyUsersInHierarchyOnly`
- `notifyParentOrganisationUnitOnly`

Anyway, I hope you think this solution is OK. Otherwise I am afraid you will have to work on it yourself, because I am off tomorrow.

PR #525, opened 2018-07-10
DHIS2-4145 relationship type

Birkbjo:
Add support for multiple fields within constraints.

The rules are like this:
TEI => TET (required) Program (Optional)
PI => Program (required)
PSI => Program (mutually-exclusive/required) ProgramStage (Mutually exclusive/required)

Note that the server does not validate required fields... So normal field validation do not work.


Birkbjo:
The "mutex" field in the options could probably be a boolean, as of right know I just check for the existence of it... However, I was thinking that it might be expanded and fields within the same "modelType" may be mutually exclusive with just another field, and not all of the fields within that modelType, even though that's the case as of right know.

Birkbjo:
Thanks for the review!

I think combining the lists are confusing, it should be obvious what kind of object you have selected.
You have to either select program or programStage for PSI (not both, not none). I thought about disabling the other if one was selected to make it clearer. However, then the user must select  "No value " to enable the other field and switch to the other object - making the selectable field nullable even though one of them must be selected. 

I might add them on the same horizontal  line with some kind of divider between, if that makes more sense. But it's kind of important to get it in soon, so I believe the current solution is good enough for now.

src/config/field-overrides/relationshipType.js
```diff
@@ -80,7 +99,28 @@ class Constraint extends Component {
...
+        const entity = this.state.relationshipEntity;
+        const modelTypes = modelTypesForRelationshipEntity[entity];
+        return modelTypes.map(objOpts => {
+            const modelType = objOpts.modelType;
+            return (
+                <DropdownAsync
+                    {...this.props}
+                    isRequired={objOpts.required}
+                    labelText={`${this.translate('select')} ${this.translate(
+                        camelCaseToUnderscores(modelType),
```
HendrikThePendric: labelText={`${this.translate('select')} ${this.translate(camelCaseToUnderscores(modelType) ${objOpts.required ? ' *' : ''}}

HendrikThePendric: This is the only change I propose for now.

PR #531, opened 2018-07-24
TECH-82/Update to new d2-ui-header-bar

Birkbjo:


varl:
Cool, nice that this works now.

PR #532, opened 2018-08-07
Handle invalid paths without throwing an error

HendrikThePendric:
- Removed error throwing from RXjs Subject, because this kills it.
- Added conditional redirect behavior to subscription function so it redirects to 'list/all' if no schema was found

In the end, the cleanest way to handle the redirect case was to include it in the subscribe handler.

### DIFF
--- a/src/List/list.actions.js
+++ b/src/List/list.actions.js
@@ -8,6 +8,7 @@ import { getInstance } from 'd2/lib/d2';
 import listStore from './list.store';
 import detailsStore from './details.store';
 import appState from '../App/appStateStore';
+import { goToRoute } from '../router-utils';
 import {
     getDefaultFiltersForType,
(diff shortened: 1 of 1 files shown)


src/List/list.actions.js
```diff
@@ -123,12 +118,19 @@ listActions.loadList
...
-
-        complete(`${schema.name} list loading`);
+        if (schema) {
+            const listResultsCollection = await getSchemaWithFilters(d2.models, schema.name)
+                .list(Object.assign(query, getQueryForSchema(schema.name)));
+    
+            listActions.setListSource(listResultsCollection);
+            complete(`${schema.name} list loading`);
+        } else {
+            // If schema is not available is is not possible to fetch a list
```
varl: `s/is is/it is`

PR #533, opened 2018-08-07
DHIS2-4369 - Fix psde assign with domainType!=TRACKER

Birkbjo:
Some program stages seem to have assigned programStageDataElements where the dataElement have domainType=AGGREGATE. This should most likely not be possible, however it crashed the app when assigning dataElements to a programStage.

I'm now checking for this before populating the store, and adding elements that are assigned to the current stage into the store - so it's possible to remove the elements.
It's still won't show up in the left side of the multi-select.

Also added a title on the name column, so titles for longer data element names may be shown on hover.

### DIFF
--- a/src/EditModel/event-program/assign-data-elements/AssignDataElements.js
+++ b/src/EditModel/event-program/assign-data-elements/AssignDataElements.js
@@ -240,17 +240,14 @@ function AssignDataElements(props, { d2 }) {
     //This is due to a database inconsistency. This fix makes it possible to show and be able to remove these
     //elements from the UI
     const otherElems = props.model.programStageDataElements
-    .filter(v => {
-        const { dataElement } = v;
-        return dataElement.domainType && dataElement.domainType !== "TRACKER";
-    })
-    .map(v => {
-        const dataElement = v.dataElement;
-        return {
+        .filter(({ dataElement }) => (
+            dataElement.domainType !== "TRACKER"
+        ))
+        .map(({ dataElement }) => ({
             id: dataElement.id,
             text: dataElement.displayName,
             value: dataElement.id,
-        }});
+        }));
 
     itemStore.setState(
         props.trackerDataElements.map(dataElement => ({
--- a/src/EditModel/event-program/assign-data-elements/AssignDataElements.js
+++ b/src/EditModel/event-program/assign-data-elements/AssignDataElements.js
@@ -199,15 +199,33 @@ const ProgramStageDataElement = pure(
     },
 );
 
-function addRenderingOptions(renderingOptions) {
-    return (psde) => {
-        const { dataElement, ...other } = psde;
+function addDisplayProperties(dataElements, renderingOptions) {
+    return ({ dataElement, ...other }) => {
+        const deDisplayProps = dataElements.find(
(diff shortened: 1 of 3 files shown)


src/EditModel/event-program/assign-data-elements/AssignDataElements.js
```diff
@@ -224,12 +236,28 @@ function AssignDataElements(props, { d2 }) {
     const itemStore = Store.create();
     const assignedItemStore = Store.create();
 
+    //Fix for DHIS2-4369 where some program stages may contain other dataelements than TRACKER
+    //This is due to a database inconsistency. This fix makes it possible to show and be able to remove these
+    //elements from the UI
+    const otherElems = props.model.programStageDataElements
+    .filter(v => {
```
HendrikThePendric: You could destructure the function parameters instead to remove a line:
`.filter({ dataElement: domainType }) => {`

src/EditModel/event-program/assign-data-elements/AssignDataElements.js
```diff
@@ -224,12 +236,28 @@ function AssignDataElements(props, { d2 }) {
...
+        const { dataElement } = v;
+        return dataElement.domainType && dataElement.domainType !== "TRACKER";
+    })
+    .map(v => {
```
HendrikThePendric: You could destructure the function parameters instead to remove a line:
`.filter({ dataElement }) => {`

src/EditModel/event-program/assign-data-elements/AssignDataElements.js
```diff
@@ -224,12 +236,28 @@ function AssignDataElements(props, { d2 }) {
     const itemStore = Store.create();
     const assignedItemStore = Store.create();
 
+    //Fix for DHIS2-4369 where some program stages may contain other dataelements than TRACKER
+    //This is due to a database inconsistency. This fix makes it possible to show and be able to remove these
+    //elements from the UI
+    const otherElems = props.model.programStageDataElements
+    .filter(v => {
+        const { dataElement } = v;
+        return dataElement.domainType && dataElement.domainType !== "TRACKER";
```
HendrikThePendric: Maybe I am missing something here, but this condition does not seem to match the comment above completely:
- From the comment I conclude that you want exclude all dataElements with domainType TRACKER, and allow all others.
- In the condition you are explicitly excluding:
   a) dataElements without a domainType
   b) dataElements with domainType TRACKER
If you only want to exclude dataElements with domainType TRACKER, then you should remove the first part of the condition. `undefined !== 'TRACKER'` will evaluate to `true` and I think that is what you want.

Quite likely this will never cause an actual problem, because dataElements always have a domainType (?). But it would improve readability if you remove the condition anyway.

Birkbjo: The idea was to exclude dataElement without domainType - just as a safeguard, but it might be unnecessary. Thanks for you review.

HendrikThePendric: It could lead to unexpected behavior. Thanks for removing it.

PR #543, opened 2018-08-30
DHIS2-4529 Remove the parent node from the org unit tree

varl:
Resolves DHIS2-4168 and DHIS2-4529.


varl:
dhis-4529_back-out-parent-node

PR #548, opened 2018-09-11
Remove partially failing workaround for API bug that has been solved

HendrikThePendric:
Some context on this bug:
1. The initial problem, as reported, was that `dataSetElements` would be assigned a `categoryCombo`, even even no override was selected in the _Override the data element category combination_ dialog.
2. Upon inspection, this was done intentionally to mitigate an API bug which was present some two years ago, see the comment in the removed method `updateCategoryCombosForDataSetElements `.
3. By now the API bug has been solved, so if no override is applied, no `categoryCombo` should be assigned to a `dataSetElement`.

So simply removing the old workaround fixes the problem.

To review if the behaviour is now correct:
1. Go to _Data sets_ and edit one, I suggest using _TB/HIV (VCCT) monthly summary_ because it doesn't have too many `dataSetElements` so is easy to inspect.
2. Click the wrench icon next to _Data elements_ and select one or more overrides in the dialog. Make sure you leave at least one without an override.
3. Close the dialog, make sure your network tab in DevTools is open and cleared, and click save.
4. Inspect the `dataSetElements` property in the payload of the POST to the `metadata` endpoint. Here you should see that:
a. `dataSetElements` with an override should have a `categoryCombos` property
b. `dataSetElements` without and override should not have this property.

### DIFF
--- a/src/EditModel/event-program/assign-data-elements/AssignDataElements.js
+++ b/src/EditModel/event-program/assign-data-elements/AssignDataElements.js
@@ -235,32 +235,26 @@ function addDisplayProperties(dataElements, renderingOptions) {
 function AssignDataElements(props, { d2 }) {
     const itemStore = Store.create();
     const assignedItemStore = Store.create();
-    const dataElementIds = new Set();
-
-    const dataElements = props.trackerDataElements.map(dataElement => {
-        dataElementIds.add(dataElement.id);
-        return {
-            id: dataElement.id,
-            text: dataElement.displayName,
-            value: dataElement.id,
-        }
-    })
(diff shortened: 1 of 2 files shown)


Birkbjo:
Do you know which version this was fixed, and if backports are necessary?

HendrikThePendric:
Good point. I just discussed this with Lars. This needs to be fixed for 2.31, 2.30 and perhaps 2.29. For 2.29 I need to find out if the API bug was reason for implementing the workaround is still present. If not, I will back-port to that too.

Do you agree with the fix though?

PR #556, opened 2018-09-26
Add style field to dataSet

HendrikThePendric:
One issues completed in one line :-)
https://jira.dhis2.org/browse/DHIS2-4630
https://jira.dhis2.org/browse/DHIS2-4631


varl:
Gotta love it when it just snaps together like that. 💯

PR #557, opened 2018-09-27
enrollment coordinate -> featureType

abyot:


varl:
@abyot Is this also safe for 2.30 API?

abyot:
> @abyot Is this also safe for 2.30 API?

No, this is only for 2.31.

PR #558, opened 2018-09-28
enrollment coordinate -> featureType

abyot:


abyot:
The API change is only for 2.31

varl:
Approved. @Birkbjo or I will merge when we are ready with 2.31 branch.

PR #559, opened 2018-10-01
Adding sonarqube analysis as a travis build step

vilkg:
Sonarqube analysis will be performed as part of every build. It will not break the build if quality gate fails.


varl:
This is still relevant. :)

PR #560, opened 2018-10-05
DHIS2-4692 Relationship options

Birkbjo:
This was way more work than I originally thought it was.

The old relationships had some issues, like listing all program stages instead of first selecting program - and then the program stages within that program.

Also refactored abit, to use more "props" instead of copying everything to state. However we still need stuff in state, due to how the props are the values that should be posted to the server. More on this in comments in the code.

I needed to implement some annyoing special cases, as in the case of editing an already existing relationship we do not know the program this programStage is belongs to. And we cannot get this from the server due the dynamic structure of the fromConstraint and toConstraint embedded objects. This object may have either "trackedEntityType", "program" or "programStage", according to selected realtionshipEntity and selected program (in case of prorgramStageInstance - event or tracker). We cannot use fieldfilter, since the relationship may not have a program - it might be a trackedEntityType relation!

I also updated the code in dropdown-async, as we needed some way to get the programType of the selected program. Now we have a callback to get the options when they are loaded. QueryParamFilter also uses deep equality to check for changes - to prevent the same filters, but different references to create an infinite loop of loading (this has been solved by keeping filters in the state in other places - but I believe it makes sense to do a deep equal-check in this case). This should not be much of a performance impact, as filters are of limited size.


src/config/field-overrides/relationshipType.js
```diff
@@ -117,28 +198,83 @@ class Constraint extends Component {
...
+                ...state.selected,
+                [modelType]: value,
+            },
+        }));
+    };
+
+    hasSelectedTrackerProgram = () =>
+        !!this.state.selected &&
+        !!this.state.selected.program &&
+        this.state.selected.program.programType === 'WITH_REGISTRATION';
```
HendrikThePendric: Why are you casting these to booleans? Existing object are also thruthy...

Let's hope JavaScript gets [optional chaining](https://github.com/tc39/proposal-optional-chaining) so we don't have to write conditions like this anymore :-(

src/config/field-overrides/relationshipType.js
```diff
@@ -59,54 +67,127 @@ const styles = {
...
+        return (props.value && props.value.relationshipEntity) || null;
+    };
+
+    getSelectedIDForModelType = modelType => {
+        // We use state here, since program and programStage cannot be selected
+        // at the same time (for the actual posted value). See constructor comment
+        return (this.state.selected && this.state.selected[modelType]) || null;
+    };
+
+    fetchProgramForProgramStage = async (programStageId) => {
```
HendrikThePendric: I think this can just be a regular method, you don't have to bind it to the this scope:
`async fetchProgramForProgramStage(programStageId) {`

src/forms/form-fields/drop-down-async.js
```diff
@@ -36,8 +37,10 @@ class DropDownAsync extends Component {
             });
         }
 
-        //referenceType updated, reload
-        if (this.props.referenceType !== newProps.referenceType || this.props.queryParamFilter !== newProps.queryParamFilter) {
+        // referenceType updated, reload
+        // Deep equal queryParamFilter, as this may have a new reference each render.
+        // Should not have that many elems.
+        if (this.props.referenceType !== newProps.referenceType || !isEqual(newProps.queryParamFilter, this.props.queryParamFilter)) {
```
HendrikThePendric: As discussed on slack, you could opt for this faster but less pretty alternative to `isEqual`:
`JSON.stringify(newProps.queryParamFilter) !== JSON.stringify(this.props.queryParamFilter)`

src/forms/form-fields/drop-down-async.js
```diff
@@ -135,19 +140,22 @@ class DropDownAsync extends Component {
             .then((options) => {
                 // Behold the mother of all hacks
                 if (!this.omgLikeJustStop) {
+                 // Behold the very special hack for renaming the very special 'default' cat combo to 'None'
+                    const renamedOpts = options.map(option => Object.assign(
+                        option,
+                        option.model &&
+                        option.model.modelDefinition &&
+                        option.model.modelDefinition.name === 'categoryCombo' &&
+                        option.text === 'default'
```
HendrikThePendric: How will this work when the system is being used in another language? I mean `option.text === 'default'`.... Won't `'default'` get translated too?

If so, then you can use this instead:
`option.text === d2i.i18n.getTranslation('default')`

src/forms/form-fields/drop-down-async.js
HendrikThePendric: Actually.... Just had another idea: You mentioned on slack that `queryParamFilter` is just an array. So before you start doing expensive object comparisons, you could just check the length first. If these don't match you know to refresh, and only if the lengths match you start doing the object comparison....

src/forms/form-fields/drop-down-async.js
Birkbjo: This is a hack that was not written by me, I just moved it to be able to return the options from the promise.
Default is a special object name that is not translated as far as I know.

PR #561, opened 2018-10-08
Dhis2 966 hierarchy operations ui

HendrikThePendric:
**Don't forget to run `yarn install` before testing because this PR requires new version of `d2` and `d2-ui` to work.**

For some context, a short summary of these fixes:
- The `OrgUnitTree` in `d2-ui` now has a new property called `foreceReloadChildren` (bool) with default value `false`. With this property set to `true`, the `OrgUnitTree` will call `root.children.load` (a method on d2 ModelCollectionProperty) with parameter `forceReload` set to `true`. This is required, because the original parent node will not have unloadedData, and d2 will return stale state.
- [The fix in `d2` makes sure](https://github.com/dhis2/d2/pull/172/files) that in the `ModelCollectionProperty.load` method, the Map is first cleared before appending new items. This was required because otherwise removing items didn't work properly, and this created problems when removing a node from its parent.

Besides the fixes in the deps above, the following changes have been done:
- Reduce the number of calls to `setAppState` (see comment below)
- Introduce a `moveTargetPath` property that is used to expand the target parent (i.e. the OrgUnit on the right into which the OrgUnits from the left are moved)


src/OrganisationUnitHierarchy/OrganisationUnitHierarchy.component.js
```diff
@@ -179,42 +180,38 @@ function moveOrganisationUnit() {
...
-                    hierarchy: {
-                        ...hierarchy,
-                        isProcessing: false,
-                    },
-                });
+                // Wait till stack clears before setting isProcessing to false and triggering a rerender
+                // Without this the UI represents a stale state
+                setTimeout(() => {
+                    setHierarchyProcessingStatus(appState.state.hierarchy, false)
+                }, 0);
```
HendrikThePendric: I've tried to reduce the number of calls to `setAppState`, but for reasons unknown to me, the UI is not updating properly without this defer....

PR #565, opened 2018-10-18
Add select option set first logic to optionGroupSet

Birkbjo:
This adds a restriction that each option group should have the same option set.


Birkbjo:
Needs backport.

PR #569, opened 2018-10-24
Better appload handling. Session expiration handling

Birkbjo:
Note: this needs a new version of d2 when https://github.com/dhis2/d2/pull/176 is merged.


Birkbjo:
This probably won't be added in 2.31... As it needs new d2 version, and current d2-ui is not compatible with d2@>2.29

varl:
:(

varl:
Just close then I guess?

PR #570, opened 2018-10-30
DHIS2-4584-delete-program-stage

Birkbjo:
Updated delete-epic to call api DELETE when deleting program stage.

If we just deleted from the programStage-array an posted this to the server, they were not properly deleted.

A consequence of this is that you cannot delete the program stage if it hasn't already been saved. So creating a stage and then deleting it results in an error.
We could consider checking for 404 when deleting, and just remove it from the state if this is the case. This might be a bit less transparent to the user though, as the same thing would happen if someone else has deleted the program stage in the mean time... And it would be difficult to differentiate between "The program stage does not exist on the server because it has not been saved yet" and "the program stage does not exist due to some other error". 

I do not want to introduce a flag in the state to check if the stage has been saved to server or not. Will just cause bugs and annoyance. 

This is a pretty serious bug, as described in the issue. Should probably be backported to both 2.30 and 2.29...

### DIFF
--- a/src/EditModel/event-program/tracker-program/epics.js
+++ b/src/EditModel/event-program/tracker-program/epics.js
@@ -18,6 +18,7 @@ import {
     PROGRAM_STAGE_EDIT_SAVE,
     deleteProgramStageSuccess,
     deleteProgramStageError,
+    CONFIRM_PROGRAM_STAGE_DELETE,
 } from './program-stages/actions';
 import { getMaxSortOrder } from './program-stages/selectors';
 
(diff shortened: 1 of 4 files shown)


HendrikThePendric:
Besides the error above that needs to be fixed, I also have a question based on your initial comment. From that I gather that before we were actually removing the reference to the `programStage` from the `program`. I imagine this would lead to correct behavior for the programs but a lot of orphaned programStages in the DB.

So now we are calling delete on the programStage, I wonder what will happen to the references to that stage on a program.... Is the server doing some sort of cascading delete so that these references are removed? (I would just like to make sure that the references on the program are also deleted)

Birkbjo:
Silly me... I forgot to return the observable, as I had just refactored from returning a promise.

As far as I understood it, is that when we just delete the references, they are not delated by hibernate due to some foreign key restrictions.
When we call delete on it directly, it gets handled properly by the deletion handler. You still get errors if the object is used somewhere else, eg someone is enrolled in the program stage.

Birkbjo:
Closing due to same PR here https://github.com/dhis2/maintenance-app/pull/573, but with same commit hashes as backport branches.

PR #586, opened 2018-11-28
Featuretype override - remove unused options - DHIS2-5419

Birkbjo:


src/config/field-overrides/program.js
```diff
@@ -6,6 +6,10 @@ import PeriodTypeDropDown from '../../forms/form-fields/period-type-drop-down';
  * between these two, as they have different behavior and overrides.
  */
 
+
+// The other options from the backend is not used... yet
+export const featureTypeOverride = ['NONE', 'POINT', 'POLYGON'];
```
varl: Do we have the list of all feature types?

It would be easier to maintain a blacklist than a whitelist.

Birkbjo:
Would be more work to use a blacklist, as then we would need to replace the component to get access to props, where the original list from the server is located.

PR #590, opened 2018-12-04
Update to new filter name `identifiable` introduced in v2.31

HendrikThePendric:
Filter with name `displayName` was gone, so it was not finding the right filter string


Birkbjo:
LGTM

PR #607, opened 2019-01-11
DHIS2-5502/Prevent rendering  programstagelist before program is selected

Birkbjo:
Prevent rendering async programstage-dropdown unless a valid tracker program has been selected.
It's not enough to pass 'disabled', as this will load the options. 

Also added initial values when editing a program rule, to prevent loading every programstage instead of just for the selected program.

Also clears the programstage when program is changed. This can only happen when creating new programrule. This is to prevent this workflow: select program -> select programstage -> select another program. This could result in creating a program rule with a programstage not in the selected program, and it would be hard for the user to recognize this.

See https://jira.dhis2.org/browse/DHIS2-5502.

### DIFF
--- a/src/config/field-overrides/program-rules/programRuleProgramStage.js
+++ b/src/config/field-overrides/program-rules/programRuleProgramStage.js
@@ -26,7 +26,15 @@ class ProgramStageField extends React.Component {
         }
     }
 
+    shouldRender = () => {
+        const props = this.props;
+        const isEventProgram = props.model.program && props.model.program.programType === "WITHOUT_REGISTRATION";
+        return props.model.program && !isEventProgram;
+    }
     render() {
+        if(!this.shouldRender()) {
+            return null;
+        }
         const props = this.props;
         const disabled = !this.state.programId;
         
--- a/src/config/field-overrides/program-rules/programRuleProgramStage.js
+++ b/src/config/field-overrides/program-rules/programRuleProgramStage.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import DropdownAsync from '../../../forms/form-fields/drop-down-async';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


src/config/field-overrides/program-rules/programRuleProgramStage.js
```diff
@@ -9,31 +10,46 @@ class ProgramStageField extends React.Component {
...
                 programId: selectedProgram.id,
                 queryParamFilter
             });
+            //Clear programStage when program is changed
+            objectActions.update({fieldName: 'programStage', value: null})
         }
     }
 
-    render() {
+    shouldRender = () => {
```
HendrikThePendric: This doesn't need the overhead of binding to the instance's `this` scope as it is not used as a callback. You could just use `shouldRender() {`

src/config/field-overrides/program-rules/programRuleProgramStage.js
```diff
@@ -9,31 +10,46 @@ class ProgramStageField extends React.Component {
...
         const props = this.props;
-        const disabled = !this.state.programId;
+        const isEventProgram = props.model.program && props.model.program.programType === "WITHOUT_REGISTRATION";
+        return props.model.program && !isEventProgram;
+    }
+
```
HendrikThePendric: I think the whole thing would be slightly cleaner like this:
```javascript 
shouldRender() {
    const { program } = this.props.model;
    return !!(program && program.programType === 'WITHOUT_REGISTRATION');
}
```

HendrikThePendric: Yeah, you're right, a mistake in my suggestion. I missed the `!` you had before `isEventProgram`.
It's definitely not worth changing this, since these were mostly cosmetic suggestions.

PR #611, opened 2019-01-18
fix(optionset): reset changedOriginalFieldValues when model changes

Birkbjo:


varl:
Patch looks good to me.

One nitpick:

`//reset field values when model changes`

I can see that the code resets the field values when the model changes, what I want to know from the comment is "why" we need to do this.

Birkbjo:
The why comment has been added

HendrikThePendric:
@Birkbjo could you have a look at the changed files and confirm this is looking good now? We'll have to resolve some small conflicts to merge this into Master. I can take care of this, i've self-assigned this PR.

Birkbjo:
LGTM!

PR #612, opened 2019-01-21
fix(program notification): Save cursor position after mouse click event

Mohammer5:
Store the cursor position inside title and message input/textarea fields
when changing the cursor position with a mouse click


src/config/field-overrides/validation-notification-template/SubjectAndMessageTemplateFields.js
```diff
@@ -58,6 +58,8 @@ export default class SubjectAndMessageTemplateFields extends Component {
         const subjectChange = (event, value) => this.props.onUpdate({ fieldName: 'subjectTemplate', value });
         const messageChange = (event, value) => this.props.onUpdate({ fieldName: 'messageTemplate', value });
         const messageLabel = d2.i18n.getTranslation('message_template');
+        const setActiveSubjectTemplateField = this.setActiveField('subjectTemplate');
+        const setActiveMessageTemplateField = this.setActiveField('messageTemplate');
```
HendrikThePendric: I realise that we were already creating 2 functions inline on each render here, but this shouldn't happen. Could you move these outside of the render method and instead do one of the following:
- Change them into "class properties" (i.e. `subjectChange = (event, value) => {...`
- Or just as regular class methods, and bind them to the this scope in the constructor.
I guess, because state is also initialised as a class property, the first option is probably the best fit with this component.

PR #615, opened 2019-01-28
Dhis2 1827 fix mandatory category combination

Mohammer5:
# Problem description

When submitting the event/tracker program form (by either creating a new one or updating an existing one), not all required fields where checked (like the category combination drop down).

# Solution

I extended the epic that was triggered on the EVENT_PROGRAM_SAVE action.
Now it'll check if there are missing fields. If there are missing fields, instead of trying to save the incomplete form on the server side and closing the form, the user will stay on the form will be notified that required information is missing.

## Checking is fields are missing

I mostly copied the code that is used to determine the form fields as that's where the required attributes are found (it won't match the required attributes on the d2 model, some fields will be overriden).
There are now functions which can be used in isolation to extract the missing fields for a specific form.

# Refactoring

I needed an additional method in the FormFieldsForModel class. I did some refactoring here, to make code more declarative instead of imperative so the next person reading the code will know exactly what happens when reading the code for the first time without having to understand what the code actually does until the dev needs to change/extend the functionality.

## modelFieldConfigCreator

I deleted this file as it contained duplicate code.

## Refactoring the epics

I think this functionality has a HUGE refactoring potential but that would mean that I have to touch a lot of things and therefore might break a lot of things, so I didn't refactor anything here, I just added the logic. I'm not sure whether a Promise is the right idea or if we should use an Either-monad.

# Unit tests

I only added one test as everything else is very hard to test (especially as there's no mock of the d2/d2-ui lib).


src/EditModel/event-program/epics.js
```diff
@@ -403,37 +404,84 @@ const saveEventProgram = eventProgramStore
             .catch(err => Observable.of(saveEventProgramError(err))),
     );
 
+/**
+ * @param {Object} eventProgramStore d2-ui store
+ * @param {Object} d2 Instance of d2
+ * @return {Promise<eventProgramStore>} A promise that resolves eventProgramStore
+*/
+const checkProgramForRequiredValues = (eventProgramStore, d2) => {
```
Birkbjo: Is there a good reason this is not a simple predicate function? I think it complicates things making it a promise. 
I would've probably called it ```getMissingFieldsForProgram``` and return an  empty array if there are no missing fields, and just check for this inside the switchMap. Then you do not need the inner flatMap below either.

Mohammer5: Yeah that was my thought as well. But we'd need some kind of wrapper as the next function in the chain needs the data of the program as well, that's why I thought of an Either monad. The reason I went for a promise is that I didn't have to add functionality like a monad, but I could've used a simple object wrapper as well, I suppose..

Any arguments against me changing the return value to the following?
`{ eventProgramStore, missingFields }`

Mohammer5: @Birkbjo Please have a look at the changes I made to the epic, should be easier to understand now as well

Birkbjo: Neat, LGTM!

PR #618, opened 2019-02-04
chore/code style husky

Birkbjo:


Birkbjo:
Do not merge, code-style is renamed to cli-style.

@varl, should we run the code-style on all files in this PR?

varl:
Better to have it in the same PR as it's related.

PR #619, opened 2019-02-08
fix: ensure proper encoding for api requests

ismay:
issue: https://jira.dhis2.org/browse/DHIS2-5081

- [x] Verify metadata request is still exactly the same

### DIFF
--- a/src/EditModel/event-program/epics.js
+++ b/src/EditModel/event-program/epics.js
@@ -151,21 +151,20 @@ function loadEventProgramMetadataByProgramId(programPayload) {
         'user[id,name]',
     ].join(',');
 
+    // Tomcat 8.5 does not allow unencoded brackets in querystrings. By passing the query params
+    // as an object to d2 it will url encode all params and escape the brackets
+    const queryParams = {
+        fields: ':owner,displayName',
+        'programs:filter': `id:eq:${programId}`,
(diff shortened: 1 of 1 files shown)


src/EditModel/event-program/epics.js
```diff
@@ -449,7 +448,7 @@ export const programModelSave = action$ =>
         .switchMap(({ eventProgramStore, missingFields }) => (
             missingFields.length
                 ? createSaveEventProgramError$()
-            
+
```
varl: :1st_place_medal:

ismay: Yeah I think that that's atom stripping out trailing whitespace. It's a bit annoying. That's why I wanted to add an editorconfig to code-style, so that I can point to that as the culprit 😋

I could remove this btw. It's nice to not have trailing whitespace, but strictly speaking probably nicer to do that separately?

varl: I think it's fine to remove trailing whitespace in a commit. Larger refactoring work might warrant a separate commit though.

Mohammer5: I'd argue that this a good thing actually. Almost all editors strip trailing white spaces.
When hitting 'end' or `$` in vim, I want to go to the last visible character, not to somewhere unexpected.

ismay: Oh yeah, don't get me wrong, I like having an editorconfig on all my projects to remove trailing whitespace, adding end of file newlines, etc. It's just that I thought some might consider it messy if it's not the subject of the pr. But I'm happy that noone minds it :)

ismay:
Merged from another patch branch to v29, 30, 31 and master (see linked PRs).

PR #633, opened 2019-02-26
DHIS2-5861 - Fix colorpicker display bug

ismay:
Fixes: https://jira.dhis2.org/browse/DHIS2-5861
Related: https://github.com/dhis2/d2-ui/pull/382


ismay:
Merged as separate patch branches (see above)

PR #638, opened 2019-03-04
Feat/dhis2 5291/metadata download

Birkbjo:
Support for metadata download though the Gear-icon in list.

Should only download items that are filtered in the list.

Note that I started some refactoring of Dialogs. The List-component rendered like 6 different dialogs, so I figured I wouldn't add more, and it was time to centralize this. DialogRouter is render in the root of the app, and can be used to render any dialog (any component really), by using redux. You can also send any props to the dialog as you wish.
I've used the same design for dialogs in the App store app, and I think its pretty simple and neat - instead of having logic for opening and closing dialogs scattered all over the place.


src/List/columns/epics.js
```diff
@@ -50,44 +77,18 @@ const editColumnsForModel = (action$, store) =>
...
-                    return genericErrorMessage;
-                }
-            }
-            try {
-                const cols = getAllModelTypes(store.getState());
-                const updatedColumns = setColumnsForModel(cols, action);
-                const ret = await namespace.set(COLUMN_KEY, updatedColumns);
-
-                return {
+        .switchMap(([action, d2]) =>
```
Birkbjo: This refactor was needed since closing the ColumnConfigDialog was done through the reducer for that dialog. As this PR also added centralized dialog-handling, and this dialog previousy had its own reducer (was nested in configurableColumns-reducer), I removed this.
Now this epic also dispatches ad action for closing once it completes.

src/Dialog/DialogRouter.js
```diff
@@ -0,0 +1,44 @@
...
+    };
+
+    render() {
+        const { open, dialogType, dialogProps, ...rest } = this.props;
+        if (!open) {
+            return null;
+        }
+        const DialogType = DialogComponents[dialogType];
+        return DialogType ? (
+            <DialogType {...{ ...rest, ...dialogProps }} open={true} />
```
ismay: I was just looking through this PR to learn a bit about the maintenance-app codebase, and was wondering, could this line not be replaced by:

```
<DialogType { ...rest, ...dialogProps } open={true} />
```

Or am I overlooking something obvious?

ismay: Ah, I assume it's because the jsx spread syntax only allows you to spread a single object.

Birkbjo: It's kinda ugly, but its a tricky way to make it an expression. Would probably be cleaner with 
`const mergedProps = {... rest, ...dialogProps }` and then `{...mergedProps}`

PR #640, opened 2019-03-11
chore: remove core-resource dep

varl:


erikarenhill:
LGTM, tested it locally + building bundle.

PR #644, opened 2019-03-18
fix(variableselector): filter variables according to predicates

Birkbjo:


src/config/field-overrides/program-indicator/VariableSelector.js
```diff
@@ -9,40 +9,54 @@ config.i18n.strings.add('program_stage_id');
...
+    incident_date: isEventProgram,
+    program_stage_id: isAnalyticsTypeEnrollment,
+    program_stage_name: isAnalyticsTypeEnrollment,
+    tei_count: overSome([isFilterType, isAggregationTypeNotCount]),
+};
 
-    return programIndicatorVariables;
+const getProgramVariables = props => {
+    const variables = programIndicatorVariables;
+    const conds = Object.keys(conditionalPIVariables).filter(variable => {
```
Mohammer5: lodash's filter function accepts objects and returns an array.

PR #660, opened 2019-04-03
feat: add support for bidirectional relationships

Birkbjo:


Birkbjo:
I think that makes sense to be honest. I won't add frontend logic to clear that value if its not something someone explicity asks for. Also that could be handled server side if they wanted. As long as its not used anywhere when bidirectional is 'false', I think its fine.

HendrikThePendric:
Agreed 👍

PR #671, opened 2019-04-12
fix(data assignment table): make width of cells dynamic & enable word

Mohammer5:
I replaced the MUI Table components with simple custom components and added some basic styles.
The MUI components separated the thead and tbody in different tables, therefore the cells had to have the same fixed width.

The data element names and the labels in the table header now don't have to be in one line anymore, allowing a denser table while everything is still readable.

This table is used in:
1. Tracker Program -> Program stages -> Select a stage -> Go to 'Assign data elements'
1. Event Program -> Assign data elements


Mohammer5:
@erikarenhill What do you mean by sub components? Just the naming? Or something technically different?

erikarenhill:
> @erikarenhill What do you mean by sub components? Just the naming? Or something technically different?

Table.js could look like something like this, only exporting the Table along the other comps

```
const Table = ({ children, ...props }) => (
    <table {...props}>
        {children}
    </table>
)

Table.Body = ({ children, ...props }) => (
    <tbody {...props}>
        {children}
    </tbody>
)

Table.Row = ({ children, ...props }) => (
    <tr {...props}>
        {children}
    </tr>
)

Table.Cell = ({ children, style, ...props }) => (
    <td>
        {children}
    </td>
)

export default Table
```

Then we could write
```
import Table from './Table'


      <Table>
        <Table.Body>
          <Table.Row>
              <Table.Cell>
                Foobar
              </Table.Cell>
          </Table.Row>
        </Table.Body>
      </Table>
```

but as I said that's just my personal preference, if the Body, Row and Cell components should only be used within the Table - component they could be "scoped" by name by this way.

Birkbjo:
Nice! I was afraid we had to update the datatables on first glance when I saw the issue.
I'm not going to be too picky on the names, so they are fine for me. 

> @erikarenhill What do you mean by sub components? Just the naming? Or something technically different?

I've seen this used in libraries like semantic-ui, but they are often combined with every module being export from the root, eg `import { Table } from 'material-ui`. In that case it makes sense, as you would want to group related components. But if you have a file called 'Table" I believe it makes sense to just export them normally, as they are already grouped together.

varl:
I've used that approach and think it's nice from a usage perspective. The only drawback is that it is not possible to tree-shake unused properties.

For a **Table** where you would never _not_ want to use the sub-components **Row**, **Cell**, as they are essential.

I don't recommend it if there is a chance the sub-components will be unused as they will end up in the bundle (confirmed with Webpack, Rollup, UglifyJS, and Terser).

Mohammer5:
@erikarenhill I've updated the names / structure according to your proposition. Looks nice tbh!
@varl Not a probleme here as all Components will def. be used.

PR #672, opened 2019-04-16
feat(predictor): conditionally show category combo for predictors

HendrikThePendric:
This PR adds a form field for the outputCombo property on the predictor model. This field was implemented as a field-override component because there were some subtleties to take into account:
- Don't show all `categoryOptionCombos`, only the ones that apply to the current `dataElement`
- Only show the field if a `dataElement` actually has any `categoryOptionCombos`, other than default

I also decided to clean up the `src/config/field-overrides/predictor.js` file a bit. I've extracted some separate entities from that file and placed them in individual files in the folder `src/config/field-overrides/predictor/`. So please note that the changes to the following files don't need to be reviewed:
```
src/config/field-overrides/predictor/ExpressionDialog.js
src/config/field-overrides/predictor/ExpressionField.js
src/config/field-overrides/predictor/expressionStatusActions.js
src/config/field-overrides/predictor/expressionStatusStore.js
```

Or to put it differently, pretty much all the actual changes are in:
`src/config/field-overrides/predictor/DataElementCategoryOptionCombo.js`

To test:
1. Go to http://localhost:8081/#/edit/otherSection/predictor/add
1. Throttle your network speed to Slow 3G to behold the DropDownLoader in its full glory
1. Select "All other new" in the "Output data element" field
1. See the dropdown appear and load
1. Switch to another "Output data element" and see it reload, and hide if no categoryOptionCombos were found
1. Select a value and click save, then verify the value persisted


src/config/field-overrides/predictor/DataElementCategoryOptionCombo.js
```diff
@@ -0,0 +1,143 @@
...
+    );
+
+    // The form builder is mutating objects which causes this.props.model.output
+    // and prevProps.model.output to be equal in the componentDidUpdate hook,
+    // after the output property was updated.
+    // By keeping track of this string literal the changes are detected correctly.
+    prevOutputId = null;
+
+    onChange = event => {
+        const fakeEvent = { target: { value: { id: event.target.value } } };
```
ismay: This seems confusing. I'd expect `value` to contain the contents of an input, not an object with an `id`. I expect you're just conforming to what the onChange callback expects, but it seems strange to me.

ismay:
It looks good to me. I'm not yet that familiar with the intricacies of the maintenance-app, so maybe @Birkbjo can comment on the functional aspects. But it works for me, and other than the comment I left on the fakeevent being confusing to me, code looks good 👍.

And of course, respect for working on the maintenance app 💪

src/config/field-overrides/predictor/DataElementCategoryOptionCombo.js
HendrikThePendric: You are right on both accounts:
-  I am indeed just conforming to what the onChange callback expects
- Yes it is strange... Let's not implement a form api in this way ever again

ismay: > Yes it is strange... Let's not implement a form api in this way ever again

Sounds like a good plan 😋

PR #674, opened 2019-05-02
fix(program stage): change wording of save btn indicating that nothing is saved

Mohammer5:
The button text changed from "Save" to "Add stage" / "Update stage", depending on the scenario.
It also says "Update stage" when a stage, that was added during creation of a new tracker program, is being edited.

Furthermore, for clarification to the end user, I included a paragraph when editing a stage, to inform the user that clicking "Update stage" will not save the stage, but the program has to be saved in order to save the stage as well.

> An updated program stage is not saved until the program itself is saved. Not saving the program will discard all changes made to the program stage.

### DIFF
--- a/src/EditModel/SaveButton.component.js
+++ b/src/EditModel/SaveButton.component.js
@@ -38,7 +38,7 @@ SaveButton.propTypes = {
 };
 
 SaveButton.defaultProps = {
-    label: undefined,
+    label: '',
     isSaving: false,
     isValid: true,
 };
--- a/src/EditModel/SaveButton.component.js
+++ b/src/EditModel/SaveButton.component.js
@@ -6,22 +6,41 @@ import addD2Context from 'd2-ui/lib/component-helpers/addD2Context';
 
 function SaveButton(props, { d2 }) {
     const {
+        label,
         isSaving,
         isValid,
         onClick,
         ...rest
     } = props;
-    const buttonText = isSaving ? d2.i18n.getTranslation('saving') : d2.i18n.getTranslation('save');
+
+    const buttonText = label
+        ? label
+        : isSaving 
+            ? d2.i18n.getTranslation('saving') 
+            : d2.i18n.getTranslation('save');
+
     return (
-        <RaisedButton {...rest} primary onClick={onClick} label={buttonText} disabled={isSaving} />
+        <RaisedButton
+            {...rest}
+            primary
+            onClick={onClick}
+            label={buttonText}
+            disabled={isSaving}
+        />
     );
 }
 
 SaveButton.propTypes = {
+    label: PropTypes.string,
     isSaving: PropTypes.bool,
     isValid: PropTypes.bool,
     onClick: PropTypes.func.isRequired,
 };
-SaveButton.defaultProps = { isSaving: false, isValid: true };
+
+SaveButton.defaultProps = {
+    label: undefined,
+    isSaving: false,
+    isValid: true,
+};
 
 export default addD2Context(SaveButton);
--- a/src/EditModel/event-program/tracker-program/program-stages/EditProgramStage.js
+++ b/src/EditModel/event-program/tracker-program/program-stages/EditProgramStage.js
@@ -1,21 +1,25 @@
-import React from 'react';
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import { compose, lifecycle } from 'recompose';
 
-import ProgramStageStepper from './ProgramStageStepper';
+import { getInstance } from 'd2/lib/d2';
 import { withProgramStageFromProgramStage$ } from './utils';
 import { changeStepperDisabledState } from '../../actions';
 import { saveProgramStageEdit, cancelProgramStageEdit, editProgramStageReset } from './actions';
-import FormActionButtons from '../../../FormActionButtons';
+import ProgramStageStepper from './ProgramStageStepper';
+import SaveButton from '../../../SaveButton.component';
+import CancelButton from '../../../CancelButton.component';
 
-const EditProgramStage = (props) => {
+const EditProgramStage = (props, context) => {
     const styles = {
         buttons: {
             padding: '2rem 1rem 1rem',
             marginLeft: '10px',
         },
     };
+
     return (
         <div>
             <ProgramStageStepper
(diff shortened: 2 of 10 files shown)


src/EditModel/event-program/tracker-program/program-stages/EditProgramStage.js
```diff
@@ -1,35 +1,61 @@
-import React from 'react';
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import { compose, lifecycle } from 'recompose';
 
-import ProgramStageStepper from './ProgramStageStepper';
+import { getInstance } from 'd2/lib/d2';
```
HendrikThePendric: I don't think this import is being used?

src/EditModel/SaveButton.component.js
```diff
@@ -6,22 +6,41 @@ import addD2Context from 'd2-ui/lib/component-helpers/addD2Context';
...
+    label: PropTypes.string,
     isSaving: PropTypes.bool,
     isValid: PropTypes.bool,
     onClick: PropTypes.func.isRequired,
 };
-SaveButton.defaultProps = { isSaving: false, isValid: true };
+
+SaveButton.defaultProps = {
+    label: undefined,
+    isSaving: false,
```
varl: You don't need to say that `label` is `undefined` or that `isSaving` is `false` by default, they will be false-y if omitted.

src/EditModel/event-program/tracker-program/program-stages/actions.js
```diff
@@ -27,9 +27,9 @@ export const changeStep = stepKey => ({
 export const nextStep = () => ({ type: PROGRAM_STAGE_STEP_NEXT });
 export const previousStep = () => ({ type: PROGRAM_STAGE_STEP_PREVIOUS });
 
-export const editProgramStage = stageId => ({
+export const editProgramStage = (stageId, addNewStage = false) => ({
```
varl: This is confusing me.

So if I call `editProgramStage(id, true)` then I am adding a new stage? And if I call `editProgramStage(id, false)` I am editing an existing program stage?

src/EditModel/SaveButton.component.js
Mohammer5: Actually I wanted to set label to empty string.. not undefined.. good that you brought this one up!
I think it's generally better to define default values so we're type safe. Most of the time it probably won't matter, but there are cases where this might lead to problems (also with primitives, like when `.toString` or `.valueOf` is used somewhere down the component tree)..
And as other devs who are working on a component down the tree probably won't check that, I think it's generally a better way to simply define everything that's optional.
What do you think about that?

src/EditModel/event-program/tracker-program/program-stages/actions.js
Mohammer5: Exactly. As the function is called `edit`, I thought that the default value should reflect the `edit` state.

You remember correctly, the maintenance app doesn't draw a distinction between editing and adding, I don't like that either. There is a PROGRAM_STAGE_ADD action though, which is used in an epic to create a model with an id and some other things, which then is used to dispatch an edit action to edit that newly created model.

I agree it'd be better to have a proper distinction here... But it seemed to be quite a hassle for a simple wording change.

src/EditModel/event-program/tracker-program/program-stages/EditProgramStage.js
Mohammer5: Good catch

Mohammer5:
@HendrikThePendric I just found out that the model is inside an observable.. There's already a helper to get that as property.. but IMO it's just a terrible idea to mix rxjs and react that way..

We just talked about this in slack and agreed that my current solution fits the purpose right now.

src/EditModel/SaveButton.component.js
varl: I just try to avoid the redundancy of `Component.defaultProps` when `Component.propTypes` does a better job.

If `SaveButton.propTypes.label` is set to `string.isRequired` then you don't need to provide a `defaultProps.label = ''` since `<SaveButton label=""/>` will satisfy the prop type check.

An empty string is not really a _good default_ value anyway, so at that point might as well just set it to required and let the consumer pass in an empty string if it absolutely must do something so silly.

Setting false-y booleans in `defaultProps` adds a lot of noise to debugging tools since all of the `defaultProps` will be added to the vDOM, e.g. for a button:

```jsx
// with defaultProps set
<Button primary={true} secondary={false} small={true} large={false} destructive={false} dense={false} />

// without defaultProps set
<Button primary small />
```

This adds more noise to the vDOM (and Storybook for component libs) which makes it harder to get an overview of what is _actually_ passed in as a prop and what is _actually_ a default value, and I tend toward the more readable option.

Sometimes `defaultProps` makes sense. In the case above I'd just do:

```jsx
SaveButton.defaultProps = {
    isValid: true,
}

SaveButton.propTypes = {
    label: propTypes.string.isRequired,
    [...]
}
```

src/EditModel/event-program/tracker-program/program-stages/actions.js
varl: Alright, yeah, might be a cost vs. benefit calculation here. :D

src/EditModel/SaveButton.component.js
varl: Just to be clear, I don't think you have to change it (other than `defaultProps.label = ''`). Just noted what I think about it.

Mohammer5: I newly introduced the `label` prop, which is just used in my case. All other instances of the `SaveButton` don't use it, because it internally determined the label. So setting it to `.isRequired` would break all other usages.

I see the point with the debug tools, I don't get what you mean by this though:
> An empty string is not really a good default value anyway

Could you elaborate what you mean by that? Because to me, the opposite has been true so far (treating empty string as no value always has worked for me)

Mohammer5: If not using an empty string, some component inside said component might do something like
`const label = props.label + ' *'`, which would result in `"undefined *"`.

Of course in this case this, for both empty string and undefined, this would be a bad case, but there are many cases where `undefined` will case worse effects than an empty string.

This is especially true when a dev works on the component which received the value, which has defaulted in the parent component. The dev probably won't even know which components are used as parents and therefore expects a string.

varl: If a component tries to do a blind operation like `const label = props.label + ' *'` the `label` prop should absolutely be set to `.isRequired`. I think that's a better way to ensure correct behaviour and have concise code. Using a default value here just seems like a convenience method over:

```
const label = (props.label || '').concat(' *')
```

Whereas actually requiring `props.label` to be a string is a more "type safe" way to deal with it.

>Could you elaborate what you mean by that? Because to me, the opposite has been true so far (treating empty string as no value always has worked for me)

Yes, it works, no disagreement there.

What I mean is that I don't think an empty string is a _meaningful_ default value. As an example, `label = 'Save'` is a more meaningful default for a string property than `label = ''`.

Mohammer5: Thanks for the explanation! I agree that empty string should be the last resort when it comes to the default value!

PR #681, opened 2019-05-07
DHIS2-2665 unset output combo when DE changes or no value is selected

HendrikThePendric:
This is a fix on https://github.com/dhis2/maintenance-app/pull/672 which had already been merged. This PR addresses 2 problems:
1. When selecting "No value" in the dropdown, I was calling `this.props.onChange` with this as a value `{ id: null }`. This was wrong, it should have been `null`. This problem would result in an error when clicking save after selecting "No value"
1. The list of "Output category combos" is only relevant in the context of a particular dataElement. As such, the `outputCombo` needs to be cleared whenever the selected dataElement changes. I have added that step now.

### DIFF
--- a/src/config/field-overrides/predictor/DataElementCategoryOptionCombo.js
+++ b/src/config/field-overrides/predictor/DataElementCategoryOptionCombo.js
@@ -42,18 +42,25 @@ class DataElementCategoryOptionCombo extends Component {
     };
 
     componentDidMount() {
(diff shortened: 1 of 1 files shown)


ismay:
I tried saving after selecting <no value>, but I'm getting this error:

Screenshot 2019-05-08 at 09 40 11

I don't know if my test was proper. Should this happen?

PR #683, opened 2019-05-08
fix: add missing width to data assignment table

Mohammer5:
Relates to DHIS2-5964


Mohammer5:
"Self approval is the most important approval. Merging."

PR #688, opened 2019-05-14
refactor: remove short name fields that are not persisted in the databae

Mohammer5:
Remove from: Data Element Group Set, Category & Organisation Unit Group Set
Fixes DHIS2-6362


Mohammer5:
Wrong branch name...

PR #695, opened 2019-05-27
fix/DHIS2-6981: orgunit geometry

Birkbjo:
The format of coordinates within Orgunits (and orgunit groups - but this does not seem to be present) has been changed to follow geoJSON structure. FeatureType is embedded within this object. Note that we still just support simple point 'features' in the app. A map picker should be supported down the line ([DHIS2-391](https://jira.dhis2.org/browse/DHIS2-391?jql=component%20in%20(EMPTY%2C%20%22%5BApp%5D%20Maintenance%22)%20AND%20text%20~%20%22map%20picker%22), surprisingly old issue!). Currenlty other featureTypes are imported, instead of selected through the UI.


src/forms/form-fields/geometry-field.js
```diff
@@ -0,0 +1,172 @@
...
+
+class GeometryField extends React.Component {
+    constructor(props, context) {
+        super(props, context);
+        this.state = {
+            longErr: undefined,
+            latErr: undefined,
+        };
+
+        this.getTranslation = context.d2.i18n.getTranslation.bind(
```
Mohammer5: No need for a constructor here

```jsx
class GeometryField extends React.Component {
    state = {
        longErr: undefined,
        latErr: undefined,
    }

    getTranslation = (...args) => this.context.d2.i18n.getTranslation(...args)
    ...
```

You could use `.bind` instead of `...args` of course

src/forms/form-fields/geometry-field.js
```diff
@@ -0,0 +1,172 @@
...
+
+    getLongLat = () => [this.getLongitude(), this.getLatitude()];
+
+    updateGeometryPoint(long, lat) {
+        const coordinates = [long, lat];
+        const geometry = {
+            type: featureTypes.POINT,
+            coordinates,
+        };
+        if (!long && !lat) {
```
Mohammer5: Shouldn't this be `||`? What happens when just one isn't given? One of them will be empty string until both values have been edited.

src/forms/form-fields/geometry-field.js
```diff
@@ -0,0 +1,172 @@
...
+                    value: geometry,
+                },
+            });
+        }
+    }
+
+    handleLatitude = event => {
+        let lat = event.target.value;
+        const long = this.getLongitude();
+        if (!isValidLatitude(lat)) {
```
Mohammer5: It's nitpicking, but there's an empty line missing here (there's an empty line in `handleLongitude`) for consistency and readability reasons. (Empty line could also be added after the else block)

But this is really just nitpicking..

src/forms/form-fields/geometry-field.js
```diff
@@ -0,0 +1,172 @@
...
+            });
+        } else {
+            this.setState({ longError: undefined });
+        }
+        this.updateGeometryPoint(long, lat);
+    };
+
+    renderLatLongFields() {
+        return (
+            <div>
```
Mohammer5: Could be `Fragment` to reduce the dom tree depth

Birkbjo: Its react 15. Does not support Fragments

src/forms/form-fields/geometry-field.js
Birkbjo: This is just used to clear the field if both are falsy. If one are undefined, you won't be able to save, due to validation.

PR #704, opened 2019-07-01
feat(expr validity): use new endpoints for indicator, validationRule & predictor

Mohammer5:
This will be ready once d2-ui supports it


Mohammer5:
@HendrikThePendric d2-ui has been updated.
I recommend testing against play/dev because it has the new endpoints

PR #707, opened 2019-07-16
fix: backport feature type override fix to v30

HendrikThePendric:
I am confused by what I am seeing here:
- For DHIS2-5419 I have found a PR against master https://github.com/dhis2/maintenance-app/pull/586 and against v31 https://github.com/dhis2/maintenance-app/pull/585 but not against v30
- So I created a new branch from v30 and cherry picked the changes from commit 1839ec0d981d626196381a50deb98a47ca10021f (which was the only commit in https://github.com/dhis2/maintenance-app/pull/586)
- And now I expected to see the same changes in 4 files as I saw in https://github.com/dhis2/maintenance-app/pull/586, but only a subset of these is showing up here...

So now my question is: 
a) Was the rest of this code already merged into v30 somehow, and were these two changes omitted deliberately?
b) Shall we proceed with this PR and merge these two changes into v30 too?

### DIFF
--- a/src/config/field-overrides/eventProgramStage.js
+++ b/src/config/field-overrides/eventProgramStage.js
@@ -9,4 +9,12 @@ export default new Map([
             },
         },
     ],
+    [
+        'featureType',
+        {
+            fieldOptions: {
+                options: featureTypeOverride,
+            },
+        },
+    ],
 ]);
--- a/src/config/field-overrides/trackedEntityType.js
+++ b/src/config/field-overrides/trackedEntityType.js
@@ -1,7 +1,19 @@
 import AssignTrackedEntityTypeAttributes from './tracked-entity-type/AssignTrackedEntityTypeAttributes.component';
+import { featureTypeOverride } from './program';
 
 export default new Map([
-    ['trackedEntityTypeAttributes', {
-        component: AssignTrackedEntityTypeAttributes,
-    }],
+    [
+        'trackedEntityTypeAttributes',
+        {
+            component: AssignTrackedEntityTypeAttributes,
+        },
+    ],
+    [
+        'featureType',
+        {
+            fieldOptions: {
+                options: featureTypeOverride
+            },
+        },
+    ],
 ]);


varl:
Cherry-picking is [a scary thing](https://dhis2.github.io/guides/git-workflow#why-is-cherry-pick-a-problem-when-used-to-port-code-between-branches), so no I don't trust it. 🤔

Birkbjo:
I think this is already in v30 (see #613). If I recall correctly, featureType is not supported in v30 (backend) for trackedEntityType. And as @varl mentions the cherry-picking results in a duplicate block: https://github.com/dhis2/maintenance-app/blob/223cd083ec949697fc91030a3315a7ad10eb9d44/src/config/field-overrides/eventProgramStage.js. As far as I can see, this is correctly backported (correct featureType for event-programs and program), and is just a stale Jira-issue.

Mohammer5:
Just a mention on the cherry pick thing here.

Afaik cherry-picking is fine if we pick from a branch that will never be merged (please correct me if I'm wrong!). The [article](https://devblogs.microsoft.com/oldnewthing/?p=98215) mentioned by @varl is mainly talking about a project with one final branch that everything will be merged into (master), but that's not our case, we have 4 master branches. I assume our structure is what the author has never considered so far, at least that's the way the first paragraph in the article seems to hint at (`Cherry-picking is a common operation in git, and it’s not a good idea. Sometimes it’s a neutral idea, but I haven’t yet found a case where it’s actually good.`). All his examples show one master branch.

So branching off of `v30`, and then cherry-picking something from `v33` (master) is totally fine and won't result in these issues, because `v33` will never be merged into the feature branch and because the feature branch for `v30` will be squashed anyways, it will result in the same as a regular commit/change/manual copying.

Cherry-picking from branch `v30` or any other branch that will eventually be merge into `v30` is really bad though and will most likely result in problems as every feature branch owner needs to know the history of merges and cherry-picks.

Looking at what I see here, it looks like @HendrikThePendric cherry-picked a commit that's been introduced by @Birkbjo which ended up on `v30`. This PR will be merged into `v30` also, so we have duplicate code in a file.

varl:
No, the same problems apply and I think that all arguments in the articles make sense in our workflow as well. In his examples we can see each of our version master as a feature branch, and our master version as his master. In article 4 he talks about:

- "What if I have multiple branches that I need to fix?"
- "What if I need to share multiple changes between the branches?"
- "What if I later realize I need to merge another fix?"

These topics correlate to our workflow. The last article (5) also has interesting use cases that apply to us.

Even if we consider we have 4 master branches and they will never be merged it is still unsafe to cherry pick between them. What he says is for a single master, but I think that's for simplicity. I'd argue it becomes even more complicated and prone to failure in our case.

On July 23, 2019 9:30:24 AM GMT+02:00, Jan-Gerke Salomon <notifications@github.com> wrote:
>Just a mention on the cherry pick thing here.>
>>
>Afaik cherry-picking is fine if we pick from a branch that will never
>be merged (please correct me if I'm wrong!). The
>[article](https://devblogs.microsoft.com/oldnewthing/?p=98215)
>mentioned by @varl is mainly talking about a project with one final
>branch that everything will be merged into (master), but that's not our
>case, we have 4 master branches. I assume our structure is what the
>author has never considered so far, at least that's the way the first
>paragraph in the article seems to hint at (`Cherry-picking is a common
>operation in git, and it’s not a good idea. Sometimes it’s a neutral
>idea, but I haven’t yet found a case where it’s actually good.`). All
>his examples show one master branch.>
>>
>So branching off of `v30`, and then cherry-picking something from `v33`
>(master) is totally fine and won't result in these issues, because
>`v33` will never be merged into the feature branch and because the
>feature branch for `v30` will be squashed anyways, it will result in
>the same as a regular commit/change/manual copying.>
>>
>Cherry-picking from branch `v30` or any other branch that will
>eventually be merge into `v30` is really bad though and will most
>likely result in problems as every feature branch owner needs to know
>the history of merges and cherry-picks.>
>>
>Looking at what I see here, it looks like @HendrikThePendric
>cherry-picked a commit that's been introduced by @Birkbjo which ended
>up on `v30`. This PR will be merged into `v30` also, so we have
>duplicate code in a file.>
>>
>-- >
>You are receiving this because you were mentioned.>
>Reply to this email directly or view it on GitHub:>
>https://github.com/dhis2/maintenance-app/pull/707#issuecomment-514089352

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Mohammer5:
> No, the same problems apply and I think that all arguments in the articles make sense in our workflow as well. 

I don't see this actually. All the conflicts he's talking about happen when a merge happens.
But the different version branches will never be merge into one another.

So what can happen:
1. We cherry pick something from `v33` into a feature branch for `v30`.
1. Another PR with changes to the same lines that are affected by the feature branch gets merged into `v30`
1. We can't merge the feature branch (with a PR) into `v30` as it's not up to date. We'll have to merge `v30` into the feature branch first. We'll have a merge conflict, which we'd have also if we applied the changes manually instead of having cherry picked.

The reason why we're safe here is:
1. We always cherry-pick exactly 1 commit from master
1. We do not add anything else to the lower-version feature branch except for fixed because master and the lower version might have diverged
1. We don't branch off of the backport branches
1. Backports will be merged with the squash & merge option, making it appear as a regular, single commit

Of course if these rules above are violated, no guarantee everything will be ok.
But so far this has always worked for me. Of course there might be manual work after cherry picking to address changes of the diverged version branches, but this would need to be done anyways.
There might be conflicts when cherry picking, but these need to be resolved by the person who knows exactly what to do.

I'm not saying we should ever cherry pick between different feature branches.
Obviously the branch, another feature branch eventually wants to be merged into, should be merged into the feature branch to get all the updates.

The only exception to me is cherry picking a specific command from `master` to a feature branch branched off a lower version branch. We should never cherry pick a commit right onto a version branch because that might cause problems. **We should only Cherry-pick up, only down**, that way I can't see this going wrong.

Anyways, the only possibility left is to apply the changes manually, if we don't want to use cherry-pick at all, because we can't merge.

varl:
I see your point, and normally it is "fine" to follow The Rules™️. I would prefer that we didn't have to backport aggressively, and if we could, have a simple workflow for it (`cherry-pick` or `merge` doesn't really matter to me).

> But so far this has always worked for me. Of course there might be manual work after cherry picking to address changes of the diverged version branches, but this would need to be done anyways.

How can you be entirely sure if there was no conflict, and the code change is not attached to a commit?

We would have to do a full manual source diff between source and target after the `cherry-pick` to be entirely sure. This is exactly what we have had to do in the past.

> I ran into a fun problem with an application today which has used cherry-picking between branches to backport, where there are "invisible" changes.
> 
> There are commits after one another in the log (call them `A` and `B`), and if I look at the patches neither of the two changed `file1` (only other changes), but when i look at the trees, `file1` contains significant changes.
>
> But the commit which made the changes doesn't exist _anywhere_.

This was from October 2018, and it caused major problems to our releases since every other build the backend was built with the wrong application version bundled, e.g. version 2.30 of the app bundled in a 2.29 build. Next build got 2.29 in a 2.29 build so it worked. Then it broke, then it worked, etc.

> There might be conflicts when cherry picking, but these need to be resolved by the person who knows exactly what to do.

This is where it has gotten dangerous. I've mentioned a situation where we wound up with a code change that was not attached to a commit, which was a pretty bad situation because:

1. A full manual diff of the source trees was required to spot the error.
2. No Git tool (`diff`, `reflog`, `log`) was able to "see" the change.
3. **No conflict happened** at time of `cherry-pick`.
4. The error was spotted **half a year after the fact**.

How is the developer supposed to know that something Bad™️ has happened if there is no conflict?

I think the problem stems from that we don't have "down" or "up" direction in our version branches. We do in a versioning sense, but the way we treat them, all our version branches are actually siblings of `master`. Naming the `master` branch `2.33` would be a more accurate way to represent the way our branches work.

Since branches exist side-by-side and independently of one-another, each version branch a conceptual fork, the problem where we have code changes that are applied without a commit attached to them can happen even if you follow "up only" or "down only" approach since we move laterally, not vertically.

I am not convinced that `cherry-pick` is a good tool for our workflow. `merge` is also not pain-free and blows up harder on tough back-ports (a good thing in a complicated situation), so it is harder not to accidentally introduce invisible (to Git) changes to the code base.

Generally the safest alternative is to re-implement the backport on each branch individually, but that is asking a Hell of a lot from developers.

The workflow you propose _should_ be safe, but since we have experienced problems without understanding how the heck Git could lose track of a code change, and given the fact that it took 6 months to find it makes we want us to err on the side of caution until we are absolutely sure and can verify it.

Mohammer5:
I see.. That's very odd behavior! I tried to find out what could cause that but I haven't found anything in regards to that..

varl:
@Mohammer5 Agreed, it is _extremely_ strange!

When I researched it I found that other people have seen the same behaviour when using `cherry-pick` and `rebase` (separately) so it seems that those two commands are more susceptible to that edge case than others. I'll see if I can retrace my steps and find the resources again.

I would wager that most can go through an entire career using those commands and not see the "invisible code change". I've used Git for around 10 years and I've _never_ seen it before.

Personally I think it is related to our workflow but until we understand exactly why it happens I want us to play it safe.

Having code changes that are invisible to the version control system is _extremely_ frightening. It's our life-line! 😰

Birkbjo:
> Generally the safest alternative is to re-implement the backport on each branch individually, but that is asking a Hell of a lot from developers.

In simple backports I've found `git format-patch` and `git am` to work well. It fails if there's a conflict though, but you can tell it to merge if you want to. To my understanding this shouldn't introduce invisible commits.

HendrikThePendric:
@Birkbjo What's going to happen with this one? Shall it be closed?

Philip-Larsen-Donnelly:
The issue has been tested successfully by Geetha on 2.30 (without this PR), so I'm closing this!

PR #712, opened 2019-08-01
fix(program data entry form): ignore CKEditor in basic & section tabs

Mohammer5:
Fixes [DHIS2-7301](https://jira.dhis2.org/browse/DHIS2-7301)

I've refactored the `getMetaDataToSend` function so I could debug the states in between.
The old function mutates the `payload` object, so in the console all logs in between were the same.
I just kept it because I think it's cleaner anyways.


src/EditModel/event-program/create-data-entry-form/CreateDataEntryForm.component.js
```diff
@@ -91,9 +106,16 @@ class CreateDataEntryForm extends Component {
                         />
                     )}
 
+                    {/* Super hacky to use the number 2 here */ ''}
+                    {/* I just didn't see another way */ ''}
```
Birkbjo: Just an idea to make it less 'hardcoded' would be to map Basic, Section, and CustomForm to a number instead of just setting `sectionFormIndex = 1`. So you could do something like`this.state.curTab === formIndex.basic`.

PR #722, opened 2019-08-06
fix(coordinatefield): support prevent saving

Birkbjo:
V30 Backport of #721


Mohammer5:
@Birkbjo This doesn't work for me
I checked out this branch, did `yarn install`, used `play.dhis2.org/2.30` as backend.
After changing nothing but either lat or long, it goes back to the list view stating "No changes to be saved" in the snackbar

Birkbjo:
@Mohammer5, sorry, I did not test on this branch. For some reason the changes in coordinate-field.js was not included. I think I just used the latest commit when running `git format-patch`. Added that commit now.

PR #737, opened 2019-09-25
feat: nextScheduleDate-field

Birkbjo:
Support for nextScheduleDate-field. See [DHIS2-6268](https://jira.dhis2.org/browse/DHIS2-6268)

I went for using a plain DropDown. Could have used DropDownAsync, but as we already have the data it's redudant, and you if you add a new DataElement to the stage this would not be included if you went back to check the dropdown, since the data is not posted until you save the entire program.

The code for the dropdown-component is messy... This seems like the first time we use the dropdown for object-values, so I changed the "key" of the rendered options to be the text of the option instead of the value. Text should be unique within a dropdown anyway.

### DIFF
--- a/src/config/field-overrides/programStage.js
+++ b/src/config/field-overrides/programStage.js
@@ -37,12 +37,13 @@ const NextScheduleDateField = props => {
             value: psde.dataElement,
         }));
 
-    let value = undefined;
-    if (props.model.nextScheduleDate) {
-        value = dataElementsOptions
-            .filter(opts => opts.value.id === props.model.nextScheduleDate.id)
-            .map(v => v.value)[0];
-    }
+    const selectedDEO =
+        props.model.nextScheduleDate &&
+        dataElementsOptions.find(
+            opts => opts.value.id === props.model.nextScheduleDate.id
+        );
+
+    const value = selectedDEO && selectedDEO.value;
 
     return <DropDown {...props} options={dataElementsOptions} value={value} />;
 };
--- a/src/config/field-overrides/programStage.js
+++ b/src/config/field-overrides/programStage.js
@@ -25,8 +25,27 @@ const reportDateOptions = [
  *  or translateOptions through fieldOptions to the DropDown-component.
  */
 
-const ReportDateToUseDropDown = props =>
-    <DropDown {...props} options={reportDateOptions} translateOptions />;
+const ReportDateToUseDropDown = props => (
+    <DropDown {...props} options={reportDateOptions} translateOptions />
+);
+
+const NextScheduleDateField = props => {
+    const dataElementsOptions = props.model.programStageDataElements
+        .filter(psde => psde.dataElement.valueType === 'DATE')
+        .map(psde => ({
+            text: psde.dataElement.displayName,
+            value: psde.dataElement,
+        }));
+
+    let value = undefined;
+    if (props.model.nextScheduleDate) {
+        value = dataElementsOptions
+            .filter(opts => opts.value.id === props.model.nextScheduleDate.id)
+            .map(v => v.value)[0];
+    }
+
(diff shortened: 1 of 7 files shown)


src/config/field-overrides/programStage.js
```diff
@@ -25,8 +25,27 @@ const reportDateOptions = [
...
+    <DropDown {...props} options={reportDateOptions} translateOptions />
+);
+
+const NextScheduleDateField = props => {
+    const dataElementsOptions = props.model.programStageDataElements
+        .filter(psde => psde.dataElement.valueType === 'DATE')
+        .map(psde => ({
+            text: psde.dataElement.displayName,
+            value: psde.dataElement,
+        }));
```
HendrikThePendric: perhaps you could also use the `id` property and use that as `key` when you are mapping over the `menuItems`?

src/forms/form-fields/drop-down.js
```diff
@@ -82,7 +82,7 @@ class Dropdown extends Component {
             .map(option => (
                 <MenuItem
                     primaryText={option.text}
-                    key={option.value}
+                    key={option.text}
```
HendrikThePendric: See previous comment, perhaps use `option.id`?

src/config/field-overrides/programStage.js
```diff
@@ -25,8 +25,27 @@ const reportDateOptions = [
...
+
+    let value = undefined;
+    if (props.model.nextScheduleDate) {
+        value = dataElementsOptions
+            .filter(opts => opts.value.id === props.model.nextScheduleDate.id)
+            .map(v => v.value)[0];
+    }
```
HendrikThePendric: `.find` would be a better fit here:
```javascript
value = dataElementsOptions.find(opts => opts.value.id === props.model.nextScheduleDate.id).value
```

src/config/field-overrides/programStage.js
Birkbjo: That would be a breaking change to the dropdown. Options doesn't have an ID.

src/config/field-overrides/programStage.js
Birkbjo: I used filter because then I wouldn't need an extra check to check if value exists. I like that I can just map it, instead of accessing props directly on something that could be undefined. `[][0]` results in undefined and not an error.

src/config/field-overrides/programStage.js
Birkbjo: You could of course check for an object if it has an id, and have a special case for this. However that seems unecessary, as the text should be a string and unique (two of the same thing in a dropdown shouldn't happen as that is very confusing and does not make sense).

HendrikThePendric: Thanks for looking into this. I would just leave things the way they are then.

src/config/field-overrides/programStage.js
HendrikThePendric: Yeah, true.... you'd have do do it it in two steps... In the future I'd suggest using the `.find` and using an `optional-chaining-operator`...

For now, I would still prefer to go with `.find` and use multiple constants:
```javascript
const selectedDEO =
    props.model.nextScheduleDate &&
    dataElementsOptions.find(
        opts => opts.value.id === props.model.nextScheduleDate.id
    )
const value = selectedDEO ? selectedDEO.value : undefined
```
Writing it like above is more readable and takes the same amount of lines as the original.

HendrikThePendric: Actually... 
`const value = selectedDEO && selectedDEO.value`
Would probably work too, since the value of an unfound `selectedDEO` should be `undefined`, and so would `props.model.nextScheduleDate`... And `undefined && undefined === undefined`

PR #738, opened 2019-10-22
fix(programstage): fetch attributeValues for programStage

Birkbjo:
The name of the attributes was not fetched with the values, so the mapping between value and attribute-name failed, causing initial attributes to not show in the form.


Birkbjo:
The bug affects all versions down to 2.29...

varl:
> 
> 
> The bug affects all versions down to 2.29...

Ouch. At least looks like the fix is _relatively_ simple to backport.

PR #941, opened 2019-11-27
fix(predictor): show missing-value-strategy dropdown in generator dialog

HendrikThePendric:
Most of the context for this issue [can be found in JIRA](https://jira.dhis2.org/browse/DHIS2-7147). Besides that there is one "controversial" part to this PR. I'll leave a code-comment for that.

### DIFF
--- a/src/config/field-overrides/predictor/ExpressionField.js
+++ b/src/config/field-overrides/predictor/ExpressionField.js
@@ -35,6 +35,10 @@ class ExpressionField extends Component {
         });
     };
 
+    missingValueStrategyChanged = missingValueStrategy => {
+        this.setState({ value: {...this.state.value, missingValueStrategy } } )
+    }
+
     render() {
         const props = this.props;
         const styles = {
@@ -48,6 +52,15 @@ class ExpressionField extends Component {
             },
         };
 
+        // Make sure to use values from the state once the dialogue is open
+        // As these values are not propagated to the props until handleSaveAndClose is called
+        const expressionDialogProps = this.state.open && this.state.value ? { 
+            ...props, 
+            value: {
+                ...this.state.value 
+            }
+        } : props
+
         return (
             <div style={styles.fieldWrap}>
                 <RaisedButton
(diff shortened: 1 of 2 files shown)


src/config/field-overrides/predictor/ExpressionField.js
```diff
@@ -56,11 +69,12 @@ class ExpressionField extends Component {
                 />
                 {props.errorText ? <div style={styles.errorText}>{props.errorText}</div> : null}
                 <ExpressionDialog
-                    {...props}
+                    {...expressionDialogProps}
```
HendrikThePendric: So this component is doing the following:
- In `handleOpen` the `value` prop is being "copied" to the state
- Changes in the dialog then trigger changes in `state.value`
- The `props.value` remain unchanged until `handleSaveAndClose` is called

It sort of makes sense to work like this, because it also needs to be possible to click cancel on the dialog and in that case the `props.value` will still be correct.

HOWEVER... Now comes the weird part... Before this fix, the `props` (not the `state`) were being spread onto the `ExpressionDialog`. And the result of this was that whenever the dropdown value changed, it seemed to be ignoring this and just rendered the old value from the props. So I've changed this part. My changes will also provide values from the state to the ExpressionManager.... Even though the ExpressionManager was working fine before, I still think this change is for the better. 

I suspect that the ExpressionManager has some form of internal state and treats its `descriptionValue` and `formulaValue` as "initial values". That would certainly explain why this was working fine before even though we were actually providing it with bad values...

HendrikThePendric:
@Birkbjo I've removed @ismay from the requested reviewers lists, because he would be prefer it if someone with more knowledge of the maintenance app would review this one... So it's up to you now 💪

ismay:
I'll review this PR for myself just so I get familiar with it though. Otherwise I'll never get to know it properly 😊

ismay:
As for functional testing. I've gone through the UI for Predictor configuration, and under Generator it's now possible to set the missing value strategy. Under sample skip test it's not present. Which is as discussed in the Jira issue 👍 Tested with debug/dev

src/config/field-overrides/predictor/ExpressionField.js
ismay: It does seem very complex (not your changes, the existing logic) for a rather simple UI pattern. With a custom store implementation, and forking props to state, etc. Though you could argue that in this case forking props to state is warranted (see the second tweet): 

Screenshot 2019-12-02 at 14 26 32
(https://twitter.com/dan_abramov/status/979520339968516097?lang=en)

In general improving this seems like it'd warrant way more in-depth structural changes. You changes make sense to me, and function how I'd expect them to, so they like a decent solution to me.

Birkbjo: I agree, it's in general an antipattern to copy props to state, and its done several places in the app, probably from before this was considered an antipattern. This might be a case where it's actually fine though. And as long as it works I see no reason to use time and change it now.

PR #1014, opened 2020-02-14
fix(predictors): set unique: false to prevent async validation

Birkbjo:
This was caused due to schema changes. `unique` in the schema was set to `true`, resulting in the form trying to validate an object as an unique string.

For reference, this is the commit that changed it: https://github.com/dhis2/dhis2-core/commit/5cba1832518a8c9870ffc9cec4e4639a674e92e9#diff-9724b46626b8d674f9cbca5f0c3251afR22

### DIFF
--- a/src/config/field-overrides/predictor.js
+++ b/src/config/field-overrides/predictor.js
@@ -152,6 +152,9 @@ export default new Map([
                 message: 'expression_is_required',
             },
         ],
+        // Override unique, it was set to true on the backend for an unkown reason
+        // see https://jira.dhis2.org/browse/DHIS2-8311
+        unique: false
     }],
     ['sampleSkipTest', {
         component: ExpressionField,
@@ -165,5 +168,7 @@ export default new Map([
                 message: 'expression_is_required',
             },
         ],
+        // see https://jira.dhis2.org/browse/DHIS2-8311
+        unique: false
     }],
 ]);


Birkbjo:
> In my personal opinion, before applying this fix, a few questions would need to be answered:

> If the change was unintentional, then we shouldn't have fixed it in this way, but the fix should have been applied server side
This is totally breaking Predictors, and it needs to be fixed ASAP. I don't agree with this at all, as it's not the server`s fault that we rely so heavily upon these seemingly volatile properties. I'm certain this is unintentional - as it doesn't really make sense to be unique; it's a complex object.
> What actually triggered this schema change? I know a lot of work has been done server-side regarding expression validation. Perhaps this change was related to that?

I would've liked an answer for this as well, and I've asked for this in the webapi channel, and as you would've guessed there's still no answer there.

src/config/field-overrides/predictor.js
```diff
@@ -152,6 +152,7 @@ export default new Map([
                 message: 'expression_is_required',
             },
         ],
+        unique: false
```
varl: These overrides needs to be documented in a comment with a link to the JIRA issue that explains why we are overriding this.

Birkbjo:
> So, if this change was an unintentional change on the webapi, this needs to be fixed there, not in the client. The maintenance-app is already complicated enough. If we start fixing stuf in the client that is wrong in the webapi, we are just making matters worse.

This is one of the main reasons why the maintenance app came to be this way. The backend is both inherently complex and full of inconsistencies. There's also been the expectation that "hacking it together" on the frontend is quicker and easier. I'm not saying that this is an excuse, or a valid reason to do it again but this issue is a prime example of why; it's urgent and we still have no valid response from the backend team as to why this change was made.

I also agree with @varl - we've seen that using these properties for dynamic logic causes way more trouble than it's worth. Even though you could argue that it "shouldn't be this way - it's exposed by the API - therefore it should be correct". This is in practice not the case - and we've seen time and time again this is not adhered to. It has improved somewhat, but we still see that we cannot rely on it. 
The original idea of having the forms backed by schemas makes sense, but I think everyone agrees that we should take this as a "lessons learned" for the maintenance refactor.

PR #1047, opened 2020-03-09
fix: add auto-scroll-body-content to all dialog instances (v34)

HendrikThePendric:
When you are reviewing this PR, could you also confirm that the base branch is correct? I am assuming, since we are in the soft-freeze period (AKA get last bug-fixes into v34), that I can still just use the plain v34 branch and not some v34.#.# branch.


HendrikThePendric:
I've had a look and since we just seem to have a v34 branch and no v34.#.# branch I think I can safely assume the base branch is correct. Will proceed to merge.

PR #1072, opened 2020-03-30
fix(data-entry-form): remove DataElement from sections when removed from stage

Birkbjo:


src/EditModel/event-program/assign-data-elements/epics.js
```diff
@@ -118,6 +121,32 @@ const removeDataElementFromStage = store => action$ =>
...
+            );
+
+            if (programStageSections && Object.keys(programStageSections).length > 0) {
+                state.programStageSections[
+                    programStageToEditId
+                ] = programStageSections.map(section => {
+
+                    const dataElementsRemoved = section.dataElements.filter(
+                        element => !dataElementIdsToRemove.includes(element.id)
+                    );
```
HendrikThePendric: The logic in here is completely sound, but I think I would have been able to comprehend things a bit more easily if the const `dataElementsRemoved` would have been named differently. Perhaps `cleanedDataElements`?

I realise the suggestion above is pretty subjective, so feel free to just merge this in as-is.

Birkbjo: Thanks, I actually struggled with the name here as I was quite exhausted when I did this. I think I will go with `updatedDataElements` as that seems to be used elsewhere.

PR #1073, opened 2020-03-30
feat(data-entry-form): add filter for section elements

Birkbjo:


Birkbjo:
Wrong Jira number in branch-name, opened new PR here https://github.com/dhis2/maintenance-app/pull/1598

PR #1356, opened 2020-08-17
feat(programrule): add validation to expression

Birkbjo:
Uses some of the same logic and components and ProgramIndicator validation. So should test that ProgramIndicator validation works as well, as I had to change some of the shared validation logic (pretty simple - just the parameter to createExpressionValidator -observable). 

See https://jira.dhis2.org/browse/DHIS2-9222

### DIFF
--- a/src/config/field-overrides/program-indicator/ExpressionStatusIcon.js
+++ b/src/config/field-overrides/program-indicator/ExpressionStatusIcon.js
@@ -1,3 +1,4 @@
+import React from 'react'
 import mapProps from 'recompose/mapProps';
 import branch from 'recompose/branch';
 import NotInterestedIcon from 'material-ui/svg-icons/av/not-interested';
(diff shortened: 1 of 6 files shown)


src/config/field-overrides/program-indicator/ProgramIndicatorExpression.js
```diff
@@ -60,7 +61,7 @@ const styles = {
         whiteSpace: 'nowrap',
     },
 };
-function ProgramIndicatorExpression({ d2, onChange, status, model, value: formula = '', referenceProperty, style, type }) {
+function ProgramIndicatorExpression({ d2, onChange, status, errorText, model, value: formula = '', referenceProperty, style, type }) {
```
Mohammer5, line 64: `errorText` is not being used

src/config/field-overrides/program-indicator/ExpressionStatusIcon.js
```diff
@@ -47,3 +66,15 @@ export function getBackgroundColorForExpressionStatus(status) {
 }
 
 export default ExpressionStatusIcon;
+
+export function ExpressionDescription(props) {
+    const { status } = props;
```
Mohammer5, line 71: Could be

```js
export function ExpressionDescription({ status }) {
```

Mohammer5 (changes requested):
Seems to work fine, just a couple of minor things

Mohammer5 (approved):

PR #1357, opened 2020-08-18
feat(dataset): add "open periods after co end date"-field

Birkbjo:
Needed to add a new operator (predicate) to the field-runner engine, as the value for `categoryCombo` is a model, and there's currently no way to check for equality here. In my opinion the `predicate` is also a good addition, and should probably have been used instead of custom made "operator types". It's a very simple rule that gets the value of the field and calls the `value` of the when-Rule, so that needs to be a function. 

See https://jira.dhis2.org/browse/DHIS2-9080 and https://jira.dhis2.org/browse/DHIS2-9081 for more information regarding the "open periods after co end date"-field.

### DIFF
--- a/src/config/field-rules.js
+++ b/src/config/field-rules.js
@@ -796,4 +796,17 @@ export default new Map([
             }],
         },
     ]],
+    ['dataSet', [
+        {
+            field: 'openPeriodsAfterCoEndDate',
+            when: {
+                field: 'categoryCombo',
+                operator: 'PREDICATE',
+                value: (categoryComboField) => categoryComboField && categoryComboField.name === 'default'
+            },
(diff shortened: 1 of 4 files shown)


src/config/field-rules.js
```diff
@@ -796,4 +796,17 @@ export default new Map([
             }],
         },
     ]],
+    ['dataSet', [
+        {
+            field: 'openPeriodsAfterCoEndDate',
```
Mohammer5, line 801: I wouldn't abbreviate category options, I didn't know what it actually meant until I saw the rendered field

Mohammer5: has to match the schema

PR #1367, opened 2020-09-07
fix(programrule): fix validation

Birkbjo:
Previously it was required to have the id of the programRule for validation. This made it impossible to validate new program rules, as the ids obviously do not exist before creation. The id is used to get the programRuleVariables, which is now changed to use the programId instead. Thanks @zubaira for the quick fix!

### DIFF
--- a/src/config/field-overrides/programRule.js
+++ b/src/config/field-overrides/programRule.js
@@ -15,33 +15,50 @@ import { ExpressionStatus } from './program-indicator/ExpressionStatusIcon';
 import { AnalyticsPeriodBoundaries } from './program-indicator/AnalyticsPeriodBoundaries';
 import createExpressionValidator from './program-indicator/createExpressionValidator';
 
+const setValidatorSubscription = (context, subFunc, programId) => {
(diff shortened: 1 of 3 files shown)


src/config/field-overrides/programRule.js
```diff
@@ -15,33 +15,50 @@ import { ExpressionStatus } from './program-indicator/ExpressionStatusIcon';
 import { AnalyticsPeriodBoundaries } from './program-indicator/AnalyticsPeriodBoundaries';
 import createExpressionValidator from './program-indicator/createExpressionValidator';
 
+const setValidatorSubscription = (context, subFunc, programId) => {
+    const path = `programRules/condition/description?programId=${
+        programId
+    }`;
+    const { status$, validate } = createExpressionValidator(path, subFunc);
+    context.validatorStatusSubscription = status$.subscribe(subFunc)
```
Birkbjo, line 23: I know this is a bit weird, but since it uses recompose we can't really have methods... Didn't want to migrate this to a class right now before freeze.

HendrikThePendric (approved):

PR #1421, opened 2020-10-09
fix(program-stage-section): fix pss displayName

Birkbjo:
The schema for `displayName` has changed to `writable: false`, which means that the d2-model will not add a setter for that property. That means a JS error was thrown when trying to edit that prop. DisplayName is readonly, but we are using it when for the title before it's sent to the server, so we need to have one at creation time and when editing. 
Turns out that we can just send the displayName to `d2.model.create` directly. 

Needed a small hack to update the displayName through `dataValues`, to update the name when its edited.

### DIFF
--- a/src/EditModel/epicHelpers.js
+++ b/src/EditModel/epicHelpers.js
@@ -111,7 +111,7 @@ export function createModelToEditProgramStageEpic(actionType, store, storeProp) 
                     to the server, and upon reloading the model, the server-defined displayName will be shown */
 
                     if (field === 'name') {
-                        updateRegularValue(model, 'displayName', value);
+                        model.dataValues.displayName = value
                     }
                 }
                 // Write back the state to the store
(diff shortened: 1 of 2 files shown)


ismay (commented):
So in testing this, I've tried changing the name of the `Child Programme` tracker program's `Birth` stage. I'm running this branch locally, with debug/dev as a backend. Changing the `Birth` stage's name and then clicking 'update stage' doesn't redirect (though it is logging redux actions on the console). I've tried this on the version of the maintenance app bundled with debug dev as well and the behaviour is the same.

However, on the maintenance app bundled with debug 2.35.0 updating does work as expected.

I'm not sure what the intended changes are. I understand the displayName field is read-only for 2.35+, but I would either expect this field in the UI to be read-only, or to be editable and for the save button to work as it did in the 2.35 version.

Let me know if I'm misunderstanding.

ismay (approved):
Ok, I was looking at the wrong section, the area that was changed works as expected (tested against debug/dev). Thanks for explaining @Birkbjo, looks good to merge 👍!

PR #1422, opened 2020-10-09
fix(programrule): show error

Birkbjo:
In some cases the error string may have already been extracted in the `objectAction`, this causes the resulting `firstErrorMessage` to be `undefined`, and thus just show an empty snackbar.


ismay (approved):
Looks good 👍

PR #1479, opened 2020-10-28
fix(programs): private program creation

Birkbjo:
This fixes so that programs can be created when the user has private Program-authority., along with a small fix to hide the creation button if the user does not have access to create a program or programStage.

Note that there's another related bug here - authority for private programStages does not exist. So that means a user needs to have public access for programStages when creating a program (but program can be private)! This is a backend issue that needs to be resolved, but it will probably not be backported.

### DIFF
--- a/src/EditModel/event-program/epics.js
+++ b/src/EditModel/event-program/epics.js
@@ -59,7 +59,6 @@ function loadEventProgramMetadataByProgramId(programPayload) {
     if (programId === 'add') {
         const programUid = generateUid();
         const programStageUid = generateUid();
-        const publicAccess = "rw------";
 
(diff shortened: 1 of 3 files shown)


varl:
https://jira.dhis2.org/browse/DHIS2-6191

varl (approved):
LGTM. Is this something we can backport to 2.32, 2.33, 2.34, 2.35 easily after verifying it in master?

Birkbjo:
> LGTM. Is this something we can backport to 2.32, 2.33, 2.34, 2.35 easily after verifying it in master?

Absolutely, we should test it thoroughly though!

varl (approved):

PR #1540, opened 2020-11-09
fix(data-entry-form): fix deletion of custom data entry form

Birkbjo:
See [DHIS2-5801](https://jira.dhis2.org/browse/DHIS2-5801).

Also added a minor update to the UI so that it defaults to the Custom-tab if a Custom-dataentry form is present.


HendrikThePendric (approved):
Code looks good.

I also tried verifying the behaviour but am not 100% sure if this is now correct. I can confirm that clicking save after removing the custom data entry form clears it now, so that seems good. However, in the issue it says _"It would also be nice to show a warning if a body without contents is saved."_. I don't think this is happening right now, but I am also not sure if this would even be relevant after applying this fix. So I guess we can just leave that to the QA team...

PR #1598, opened 2020-11-19
feat(data-entry-form): add filter for section elements

Birkbjo:
Adds filtering for dataElements when creating data entry form sections. Both in Tracker Programs -> Program Stages -> Create data entry form -> Section, and Event Program -> Create data entry form -> Section.

See https://jira.dhis2.org/browse/DHIS2-8558


src/EditModel/event-program/create-data-entry-form/Section.component.js
```diff
@@ -121,54 +173,43 @@ class Section extends Component {
...
-        return (
-            <div
-                style={{
-                    ...styles.sectionContainer,
-                    borderColor: this.props.selected ? grey800 : grey300,
-                }}
-            >
+        const sectionContent = elements && elements.length > 0 ? <div style={styles.sectionContent}>
+                    <SortableSectionDataList distance={4} onSortEnd={this.onSortEnd} onDataElementRemoved={this.props.onDataElementRemoved} sectionDataElements={filteredElements} />
+                    <HiddenElementsText numberOfHiddenElements={numberOfHiddenElements} getTranslation={this.getTranslation} />
```
Mohammer5, line 178: This needs some formatting, the lines are quite long

Mohammer5 (changes requested):
Implementation works as expected. Just one change request and one proposal: Searching causes the height of the data elements to shrink and the viewport jumps up, which is a bit confusing. We could assign a fixed height to the data elements container so that doesn't happen?

Mohammer5 (approved):
Looks much better! Also the visual bug when there was no search result is gone :ok_hand:

PR #1635, opened 2020-11-27
feat(relationshipType): make fromToName and toFromName translateable

Birkbjo:
See https://jira.dhis2.org/browse/DHIS2-9056


Mohammer5 (approved):
LGTM

PR #1684, opened 2020-12-07
feat(icon-field): fallback to dhis2-logo if selected icon does not exist

Birkbjo:
See https://jira.dhis2.org/browse/DHIS2-10095

There's going to be a cleanup of icons on the backend so that will probably make some selected icons fail. 
Need to double-check that the current fallback is correct.


varl (approved):
Implementation is OK.

Maybe there should be a different fallback icon added to the core which we can use later?

Birkbjo:
> Maybe there should be a different fallback icon added to the core which we can use later?

I will follow up with @stian-sandvold before merging this. Not sure the icon-paths are correct for the new icons.

PR #1783, opened 2021-01-08
fix: hierarchy operations paging

Birkbjo:
See https://jira.dhis2.org/browse/DHIS2-9860

Also added a small feature to make it easier to test this - shift clicking a parent will add all children. Note that this will only work when the parent is expanded.

### DIFF
--- a/src/OrganisationUnitHierarchy/OrganisationUnitHierarchy.component.js
+++ b/src/OrganisationUnitHierarchy/OrganisationUnitHierarchy.component.js
@@ -105,17 +105,31 @@ function onClickLeft(event, model) {
         .take(1) // Only grab the current state
         .subscribe(
             (hierarchy) => {
-                let selectedLeft = [];
-                const indexOfModelInSelected = hierarchy.selectedLeft
-                    .map(model => model.id)
-                    .indexOf(model.id);
+                const selectAllChildren = event.shiftKey
 
-                if (indexOfModelInSelected >= 0) {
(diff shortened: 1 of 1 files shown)


HendrikThePendric (approved):
Code is looking good, and the bug is fixed. Your shift-click trick works well too.

I did initially have some problems when testing this issue. I first got the impression that the functionality was broken and some random number of items was being moved. However, it turned out that I was simply being impatient. It seems we are doing an individual request for each OU we're moving, and also showing a toast/snackbar message for each one. Since these follow one another pretty quickly it's impossible to read and I just ended up refreshing the page too quickly. Anyway, the issue is fixed and the problem I am describing now is out-of-scope currently.

Birkbjo:
Thanks @HendrikThePendric !

> I did initially have some problems when testing this issue. I first got the impression that the functionality was broken and some random number of items was being moved. However, it turned out that I was simply being impatient. It seems we are doing an individual request for each OU we're moving, and also showing a toast/snackbar message for each one. Since these follow one another pretty quickly it's impossible to read and I just ended up refreshing the page too quickly. Anyway, the issue is fixed and the problem I am describing now is out-of-scope currently.

I noticed this is as well. I'm pretty sure we have to send multiple request as we have no direct endpoint to do this at once. However, I actually played around with not showing the snackbar until everything is finished, as I believe it is indeed pretty confusing as is. I can look into this further but in a seperate issue.

PR #1784, opened 2021-01-13
fix(programs): use correct rendering type for attributes with optionSet

Birkbjo:
`getRenderTypeOptions`-function looks for `optionSet` to determine the correct redering options. However the `programTrackedEntityAttribute` was passed, which does not include the optionSet-property, this property is at the `trackedEntityAttribute`. 

https://jira.dhis2.org/browse/DHIS2-10184


HendrikThePendric (approved):
This looks exactly like what you suggested on Slack yesterday. LGTM 👍

PR #1800, opened 2021-02-22
feat(list): get translatable properties from schema for all objects

Birkbjo:
https://jira.dhis2.org/browse/DHIS2-8897

Note that fields not defined in the 'field-order' for the model is not included. This is to prevent some models from showing fields to be translated that are not editable from the UI. In most cases the `field-order` for that model is a superset of translatable-keys, but sometimes a key might not have been added yet - so this prevents some weird discrepancies.

### DIFF
--- a/src/List/List.component.js
+++ b/src/List/List.component.js
@@ -101,28 +101,16 @@ function actionsThatRequireDelete(action) {
     return false;
 }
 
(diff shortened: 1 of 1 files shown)


HendrikThePendric (approved):
> Note that fields not defined in the 'field-order' for the model is not included

That seems to make a lot of sense! Plus code is looking good as well!

PR #1801, opened 2021-02-24
feat(expressions): add expression functions

Birkbjo:
Due to problems with upgrading and linking the old version of d2-ui, the ExpressionManager is copied from https://github.com/dhis2/d2-ui/blob/v29/src/expression-manager/ExpressionManager.js . Due to the closed, encapsulated structure of this, it was otherwise impossible to add the new functions to the manager.


varl:
I prefer this approach because the ExpressionManager from d2-ui is only used in the Maintenance app, so bringing it inline is preferable to spending time on updating `d2-ui@29`.

varl (approved):

PR #1812, opened 2021-03-10
fix(programs): request publicAccess explicitly

Birkbjo:
Fixes the `Access`-stage in programs. `publicAccess` is not returned when requested by `:owner` anymore, so the sharing-component was not shown at all due to missing information. 

See https://jira.dhis2.org/browse/DHIS2-10464


Mohammer5 (approved):
I have trouble running the app locally (has nothing to do with your change).
Code changes look good

HendrikThePendric (approved):

PR #1816, opened 2021-03-15
fix(sqlView): format non-primitive values

Birkbjo:
https://jira.dhis2.org/browse/DHIS2-10354

SQL-views that had cells with objects in them (eg. `Data elements` in Demo-DB) would crash the SQL-view component as React does not allow for rendering objects. I've added a simple formatter that checks for non-primitive values that will be stringified. 

Another strange this I encountered was that the `catch` in `componentDidMount` caught the error in the `render`-method that caused this. We should try to log errors when they are caught unless we know exactly what the error is.

### DIFF
--- a/src/SqlView/SqlView.component.js
+++ b/src/SqlView/SqlView.component.js
@@ -138,8 +138,8 @@ class SqlView extends Component {
 
     /* eslint-disable react/no-array-index-key */
     renderCell = (cell, index) => {
-        const formattedValue =
-            Object(cell) !== cell ? cell : JSON.stringify(cell, null, 2);
+        const isObject = typeof cell === 'object' && cell != null;
+        const formattedValue = isObject ? JSON.stringify(cell, null, 2) : cell;
 
         return <td key={`cell${index}`}>{formattedValue}</td>;
     };
--- a/src/SqlView/SqlView.component.js
+++ b/src/SqlView/SqlView.component.js
@@ -46,6 +46,7 @@ class SqlView extends Component {
             const { listGrid } = await d2.Api.getApi().get(url, queryParams);
             this.setState({ listGrid });
         } catch (error) {
+            console.error(error)
             backToList();
             snackActions.show({ message: d2.i18n.getTranslation('sql_query_error'), action: 'ok' });
         }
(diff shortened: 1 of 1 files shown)


ismay (commented):
I took a look and two things stood out to me:

- When testing with debug.dhis2.org/dev on `#/list/otherSection/sqlView`, I noticed that the fix works for `Data elements`. However, for `Data element by value type` I'm still getting the `Could not execute SQL` error when I click `Show SQL view`
- I was wondering why you're using `Object(cell) !== cell` to check if cell is an object. I'd expect `typeof cell === 'object'` there.

Other than those two questions it looks good to me 👍

Birkbjo:
> * When testing with debug.dhis2.org/dev on `#/list/otherSection/sqlView`, I noticed that the fix works for `Data elements`. However, for `Data element by value type` I'm still getting the `Could not execute SQL` error when I click `Show SQL view`

Well in that case the server request fails. Could probably show the error message in the snackbar as well. Eg that request fails due to ```SQL query contains variables not provided in request: `[valueType]```

This fix is just for the frontend-bug where some SQL-views that return a 200-ok response failed to display on the client.

> * I was wondering why you're using `Object(cell) !== cell` to check if cell is an object. I'd expect `typeof cell === 'object'` there.

That's indeed probably easier to reason about, but it's mostly due to `typeof null` evaluates to `'object'`. `Object(v) === v` is a simple way to check for a true primitive value. So in case of null, `null` is actually rendered and thus not shown in the column as opposed to showing `null`, which is less clean imo. 

But I can change it to `typeof cell === 'object' && cell != null ? JSON.stringify(cell) : cell` if you think that's cleaner.

ismay:
> > * When testing with debug.dhis2.org/dev on `#/list/otherSection/sqlView`, I noticed that the fix works for `Data elements`. However, for `Data element by value type` I'm still getting the `Could not execute SQL` error when I click `Show SQL view`
> 
> Well in that case the server request fails. Could probably show the error message in the snackbar as well. Eg that request fails due to `` SQL query contains variables not provided in request: `[valueType] ``
> 
> This fix is just for the frontend-bug where some SQL-views that return a 200-ok response failed to display on the client.

Ah I see, I didn't know that. I thought that that error had the same cause. Ok sounds good!

> > * I was wondering why you're using `Object(cell) !== cell` to check if cell is an object. I'd expect `typeof cell === 'object'` there.
> 
> That's indeed probably easier to reason about, but it's mostly due to `typeof null` evaluates to `'object'`. `Object(v) !== v` is a simple way to check for a true primitive value. So in case of null, `null` is actually rendered and thus not shown in the column as opposed to showing `null`, which is less clean imo. But you could also argue that we should show `null` instead of an empty cell, maybe to differentiate between empty string and `null`?
> 
> But I can change it to `typeof cell === 'object' && cell != null ? JSON.stringify(cell) : cell` if you think that's cleaner.

I see, didn't know that null could be an option as well. Yes, I think personally I'd find something along those lines a little more straightforward. Say like this:

```js
const isObject = typeof cell === 'object' && cell != null
const formattedValue = isObject ? JSON.stringify(cell) : cell
```

That would make it immediately obvious to me what's going on there. What do you think? I wonder if skipping the construction of a new object for each test would speed things up btw., since there are quite some cells, but I don't know tbh.

ismay:
> But you could also argue that we should show null instead of an empty cell, maybe to differentiate between empty string and null?

Actually. In thinking about it, yeah I think showing null might be a solution as well.

HendrikThePendric (approved):
I'd probably be in favour of the small tweak @ismay suggested
```js
const isObject = typeof cell === 'object' && cell != null
const formattedValue = isObject ? JSON.stringify(cell) : cell
```
But other than that I approve of the general approach.

PR #1817, opened 2021-03-19
fix(program): fix sharing settings not included in payload

Birkbjo:
Related to https://jira.dhis2.org/browse/DHIS2-10464 and #1812 

Reopening the issue as a bug was found when saving a program. The fields were not saved to the server due to being filtered out, my first fix just resolved getting the data - not actually posting them as well.

This fix includes the sharing fields that was changed to `owner: false` when generating the final object before posting.

### DIFF
--- a/src/EditModel/event-program/event-program-store/getMetaDataToSend.js
+++ b/src/EditModel/event-program/event-program-store/getMetaDataToSend.js
@@ -7,7 +7,7 @@ import {
     values,
     flatten,
 } from 'lodash/fp';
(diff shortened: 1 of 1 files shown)


HendrikThePendric (approved):

HendrikThePendric (changes requested):
I was a bit too quick approving these changes. The code looks good, but actually I'm not quite sure if what I am seeing when testing the app is correct. TBC.

HendrikThePendric (approved):
After switching to a more recent version of dhis2-core, everything worked perfectly.

PR #1827, opened 2021-03-24
fix(expressionmanager): show .periodOffset for indicators only

Birkbjo:
`.periodOffset` is only valid for indicators. This was added in #1801 

See Jims comment https://jira.dhis2.org/browse/DHIS2-7581


HendrikThePendric (approved):
Code looks good. I also tried to test the UI and this seems to be working well too:
- When editing an indicator I see the ` .periodOffset( ` 
- When editing a validation rule or program indicator I do not see it

The above seems to be the way to check if the functionality works, but I'm not 100% sure. If this isn't the way and you'd like me to confirm things are working in a different way, please contact me.

PR #1961, opened 2021-06-02
refactor(color picker): update color-picker to use local ColorPicker component

Mohammer5:
Fixes DHIS2-11227
([link](https://jira.dhis2.org/browse/DHIS2-11227))

I've updated the component in `color-picker.js` to use the local `ColorPicker` component instead of the one from d2-ui.
This file seems to be imported only by the `organisationUnitGroup.js` (which is the one that needed the fix), so I don't think that this will cause any regression.


Birkbjo (approved):
Not sure if we need the nested `style`-object, could probably just set color directly? Minor thing though, LGTM.

PR #1996, opened 2021-06-24
fix(stepper): show attributes in steppers

Birkbjo:
https://jira.dhis2.org/browse/DHIS2-7230

Adds a `isAttribute`-property to the fieldConfig to easily check if a field is an attribute. Previously the attribute fields were hidden in steppers as all fields not defined in `field-groups.js`-were hidden. 

ProgramIndicators seems to be the only model that is a stepper and supports attributes.

Note that I've hardcoded attributes to show up in the first step. This might not be ideal, but as we are looking to refactor the app, and the first step tends to be the "general" section of a stepper anyway, I think its fine. 
Another solution would be to introduce a special field in the "field-groups"-fields to check for and insert attributes where that special field occurs (eg. something like `__ATTRIBUTES__`).

### DIFF
--- a/src/EditModel/EditModelForm.component.js
+++ b/src/EditModel/EditModelForm.component.js
@@ -64,6 +64,7 @@ const modelToEditAndModelForm$ = Observable
 
                 // Check if value is an attribute
                 if (Object.keys(modelToEdit.attributes || []).indexOf(fieldConfig.name) >= 0) {
+                    fieldConfig.isAttribute = true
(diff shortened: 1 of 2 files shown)


ismay (approved):
Looks good! Seems like most flexible, practical solution 👍. Maybe we should annotate the fix with a couple of comments for posterity, just so that others will be able to understand what's going on more easily.

ismay:
Looks good! Should be good to merge this right?

PR #2007, opened 2021-06-25
feat(image-select): add image select field to org unit form DHIS2-11175

Mohammer5:
Closes DHIS2-11175
([link](https://jira.dhis2.org/browse/DHIS2-11175))

The image select works the following way:
* The image select has the following states: `loading`, `error`, `initialized`, `pending`, `hasInitialValue` & `removed`
  * `hasInitialValue` & `removed` are being used to display a text that says that the form has to be submitted in order to save the removal
  * `initialized` is being used to check whether the image is still `PENDING` or whether an error has occured (neither `PENDING` nor `STORED`)
  * `pending` causes a text to be rendered instead of an image, informing the user that the image is being processed on the server side
* When the user uploads an image, the post request to `fileResources` will always respond with `storageStatus: 'PENDING'`
  * So in order to display the image, the image select checks the `storageStatus` periodically (every second; up to 5 times) to update the `pending` status and display the image in case the status has changed. The text for the pending state just remains the same and is supposed to be meaningful even when the maximum amount of polls have been made and the status is still pending

#### OU has no image
image

#### Image is being uploaded
image

#### Image is being processed
image

#### Image form value is dirty (image has been uploaded and is pending or stored or file has been removed)
image

#### error occurred while uploading
image


Birkbjo:
Code looks good to me!
However it would be great if we could streamline the UI a little bit. Doesn't necessarily need to be identical, but would be great to move it closer to eg. how we select icons:
image

eg. change buttons to be of the same color. And I don't think we really need a `reset`-button, as that is basically done by cancelling the form. I was at least a bit unsure about what it did at first.

Mohammer5:
I didn't add the reset button either, but added it after I realised that cancelling the form will also discard all other potentially made changes. In my experience (which is very subjective of course) I always add pictures as the very last thing when editing profiles/forms and if you don't have the original on your computer, cancelling the whole form could be frustrating.
What do you think?

I'll adjust the button styles!

src/i18n/i18n_module_en.properties
```diff
@@ -2259,3 +2259,14 @@ open_periods_after_co_end_date=Open periods after category option end date
...
 add_section_to_form=Add section to form
+label_organisation_unit_image=Organisation unit image
+org_unit_image_alt_text=Organisation unit image
+org_unit_image_no_image_text=Organisation unit does not have an image
+org_unit_image_loading_image_data_text=Loading image data
+org_unit_image_uploading_image_text=Uploading image
+org_unit_image_image_upload_error_text=There is an error with the image:
+org_unit_image_image_pending_text=The image has been uploaded and is being processed.
+org_unit_image_storage_status_error=Image could not be stored, storageStatus is:
+org_unit_image_save_reminder=Please submit the form to save your changes
```
cooper-joe, line 2270: I suggest `Submit the form to save changes`

src/i18n/i18n_module_en.properties
```diff
@@ -2259,3 +2259,14 @@ open_periods_after_co_end_date=Open periods after category option end date
 element_hidden_by_filter=1 element hidden by filter
 $$total$$_elements_hidden_by_filter= $$total$$ elements hidden by filter
 add_section_to_form=Add section to form
+label_organisation_unit_image=Organisation unit image
+org_unit_image_alt_text=Organisation unit image
+org_unit_image_no_image_text=Organisation unit does not have an image
+org_unit_image_loading_image_data_text=Loading image data
```
cooper-joe, line 2265: I think we can remove `data` here and just use `Loading image`

src/i18n/i18n_module_en.properties
```diff
@@ -2259,3 +2259,14 @@ open_periods_after_co_end_date=Open periods after category option end date
...
+org_unit_image_uploading_image_text=Uploading image
+org_unit_image_image_upload_error_text=There is an error with the image:
+org_unit_image_image_pending_text=The image has been uploaded and is being processed.
```
cooper-joe, line 2268: I suggest `Upload successful, processing image`

cooper-joe (commented):
I agree with @Birkbjo that it would be good to match the button styles with the existing Maintenance app.

Regarding the `Reset` button, I think it might be best to adjust the label to `Reset changes`. This makes it clear that it will reset back to the previous state, not to an empty state. I would also suggest hiding the `Reset changes` button until changes are made.

I've also made some suggestions to the helper texts.

Mohammer5:
@Birkbjo I've changed a little bit of code, nothing wild
https://github.com/dhis2/maintenance-app/blob/c2417f9ed294652bf44a05e0c08e02a47b3a5e01/src/forms/form-fields/image-select.js#L331

---

@birk & @cooper-joe I've updated the styles, see below

#### OU has no image
image

#### Image is being uploaded
image

#### Image is being processed

##### Replacing the old image
image

##### New image
image

#### Image form value is dirty (image has been uploaded and is pending or stored or file has been removed)

##### Org unit has no image so far
image

##### Org unit had a different image
image

##### Previous image has been removed
image

#### error occurred while uploading
image

cooper-joe:
I think in the `OU has no image`, first example, it's not necessary to show the `x` icon - this could be misinterpreted as a button that has some function, or as an actual image of an `x`. I think not showing anything makes it clear there is no image.

...and just to check, is `Something went wrong  ¯\_(ツ)_/¯` actually being displayed? or is this an example error message from backend?

Mohammer5:
@cooper-joe 

> I think in the OU has no image, first example, it's not necessary to show the x icon - this could be misinterpreted as a button that has some function, or as an actual image of an x. I think not showing anything makes it clear there is no image.

I assume same argument applies to when there's an error?

> ...and just to check, is Something went wrong ¯\_(ツ)_/¯ actually being displayed? or is this an example error message from backend?

Yes, that's just an example backend error :laughing:

cooper-joe:
> I assume same argument applies to when there's an error?

Yeah, that too - didn't notice that 👍

Mohammer5:
@cooper-joe I've removed the "x" entirely, which changes 3 different states:

#### Error (This time with a real error message :laughing:):
image

#### No image:
image

#### Image has been removed:
image

Birkbjo (approved):

PR #2026, opened 2021-07-02
fix(image select): do not use the base url with api.get

Mohammer5:
Also contains the style changes that we've discussed in #2007

### DIFF
--- a/src/forms/form-fields/image-select.js
+++ b/src/forms/form-fields/image-select.js
@@ -98,7 +98,7 @@ export class ImageSelect extends Component {
 
     checkStorageStatus(id) {
         return this.api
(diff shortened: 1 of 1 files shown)


src/forms/form-fields/image-select.js
```diff
@@ -303,7 +300,7 @@ export class ImageSelect extends Component {
                     </ImageSelectButton>}
                 </div>
 
-                {hasNoImage && !error && !loading && !isPending && (
+                {hasNoImage && !this.state.removed && !error && !isPending && (
```
Birkbjo, line 303: Cleaner to destructure `removed` as well?

Birkbjo (approved):

PR #2046, opened 2021-08-25
feat(program-rule-action): validate program rule action expression

ghost:
https://jira.dhis2.org/browse/DHIS2-10983

It seems prettier hadn't been run on this file before the changes I made, so there are a fair amount of prettier changes in here. Main changes I've made are:

- Add local `status` state [here](https://github.com/dhis2/maintenance-app/pull/2046/files?file-filters%5B%5D=.js#diff-8601ce25b8b2c50a66176634d52f5bd353b06a582a119be15d66ade2d40ae5d4R86)
- Add validate method [here](https://github.com/dhis2/maintenance-app/pull/2046/files?file-filters%5B%5D=.js#diff-8601ce25b8b2c50a66176634d52f5bd353b06a582a119be15d66ade2d40ae5d4R331-R388)
- Run validate on updates to the `data` field [here](https://github.com/dhis2/maintenance-app/pull/2046/files?file-filters[]=.js#diff-8601ce25b8b2c50a66176634d52f5bd353b06a582a119be15d66ade2d40ae5d4R306-R308)
- Pass local status state to the data component (it already had logic for displaying validation status it turns out) [here](https://github.com/dhis2/maintenance-app/pull/2046/files?file-filters%5B%5D=.js#diff-8601ce25b8b2c50a66176634d52f5bd353b06a582a119be15d66ade2d40ae5d4R606)

Backend questions:

- [ ] I couldn't get the backend to fail validation for anything, so we should check that that's working as expected.
- [ ] It also seems that the backend doesn't substitute variables in the same manner as the program rule expression validation (see image below): 

Screenshot 2021-08-25 at 17 29 06

### DIFF
--- a/src/config/field-overrides/program-rules/programRuleActionDialog.component.js
+++ b/src/config/field-overrides/program-rules/programRuleActionDialog.component.js
@@ -12,9 +12,7 @@ import ProgramRuleConditionField from './programRuleConditionField.component';
 import snackActions from '../../../Snackbar/snack.actions';
 import RefreshMask from '../../../forms/form-fields/helpers/RefreshMask.component';
 import sortBy from 'lodash/fp/sortBy';
-import {
-    ExpressionStatus,
-} from '../program-indicator/ExpressionStatusIcon';
+import { ExpressionStatus } from '../program-indicator/ExpressionStatusIcon';
 
 function toDisplay(element) {
     return {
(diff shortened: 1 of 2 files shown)


src/config/field-overrides/program-rules/programRuleActionDialog.component.js
```diff
@@ -237,36 +331,113 @@ class ProgramRuleActionDialog extends React.Component {
...
+                }
+
+                this.setState(prevState => ({
+                    ...prevState,
+                    status: newStatus,
+                }));
+            }
+
+            // Rethrow if not a valid error
+            throw error;
```
Birkbjo, line 401: I think we should gracefully handle this as well, `api.post` will throw on network error, or eg. 500 errors. I think we should inform the user of that instead of rethrowing.

PR #2078, opened 2021-11-01
feat: support adding vector style map layer

jenniferarnesen:
Implements https://jira.dhis2.org/browse/DHIS2-12063
External map layers of type Vector style have slightly different data requirements than the other layer types.
The goal of this implementation was to minimize the change surface area and utilize existing properties (i.e. "layers") rather than adding a new property "beforeLayerId" (the "layer" properties are exactly that needed by "beforeLayerId")

TODO: confirm that string changes are correct.

image


Birkbjo (approved):

turban (approved):
LGTM!

PR #2124, opened 2022-01-26
feat(dataset-section): add checkbox to disable auto grouping

Birkbjo:
https://jira.dhis2.org/browse/DHIS2-11600

Implements a checkbox in DataSet section to disable auto grouping of data elements according to their categoryCombination in the Data Entry App. Note that this is not implemented in the current struts Data Entry app, but is implemented in the work-in-progress new Data Entry App (https://github.com/dhis2/data-entry-app)

Open to suggestions for the label, see comment in Jira-ticket.

On a side-note, this  component behaves totally different to the rest of the maintenance app, and every field is hardcoded... It was hard to resist the urge to start refactoring and reduce the repetition of code...


Mohammer5:
Code looks good, but Joe made some suggestions in the Jira issue

PR #2136, opened 2022-02-24
feat(program): add webhook for program notifications

Birkbjo:
[olhttps://jira.dhis2.org/browse/DHIS2-10855](https://jira.dhis2.org/browse/DHIS2-10855)

### DIFF
--- a/src/config/field-rules.js
+++ b/src/config/field-rules.js
@@ -642,6 +642,23 @@ export default new Map([
             }],
         },
         {
+            field: 'notificationRecipient',
+            when: [{
+                field: 'notificationRecipient',
+                operator: 'EQUALS',
+                value: 'WEB_HOOK',
+            }],
+            operations: [{
+                field: 'deliveryChannels',
+                type: 'CHANGE_VALUE',
+                setValue: (model, fieldConfig) => {
+                    if(fieldConfig) {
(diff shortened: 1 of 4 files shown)


src/config/field-rules.js
```diff
@@ -739,6 +756,23 @@ export default new Map([
...
+            operations: [{
+                field: 'deliveryChannels',
+                type: 'CHANGE_VALUE',
+                setValue: (model, fieldConfig) => {
+                    if(fieldConfig) {
+                        fieldConfig.value = model[fieldConfig.name] = ['HTTP']
+                    }
+                }
+            }],
+        },
```
Mohammer5, line 775: Duplicate rule, same as the one added above

Mohammer5 (changes requested):

src/config/field-rules.js
Birkbjo: Yup, but they're under different models.

Mohammer5: Ah, oops, didn't notice. Yeah I wouldn't put that into a variable either

Mohammer5 (approved):

src/config/field-rules.js
Birkbjo: Not easy to see from GitHub diffs, since that was truncated 😄

PR #2154, opened 2022-03-14
DHIS2 6276 booleans

tomzemp:
This PR updates maintenance app to display booleans: https://jira.dhis2.org/browse/DHIS2-6276

The code will also stringify data types other than null, but these should not be returned from a request to sqlViews resource.

Null was mapped to an empty string to make explicit the choice of no display for this data type.

### DIFF
--- a/src/SqlView/SqlView.component.js
+++ b/src/SqlView/SqlView.component.js
@@ -138,8 +138,9 @@ class SqlView extends Component {
 
     /* eslint-disable react/no-array-index-key */
     renderCell = (cell, index) => {
-        const isObjectOrBoolean = ['object', 'boolean'].includes(typeof cell) && cell != null;
-        const formattedValue = isObjectOrBoolean ? JSON.stringify(cell, null, 2) : cell;
+        const formattedValue =
+            cell === null ? '' :
+            (typeof cell === 'object' ? JSON.stringify(cell, null, 2) : String(cell));
 
         return <td key={`cell${index}`}>{formattedValue}</td>;
     };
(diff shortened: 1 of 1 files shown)


src/SqlView/SqlView.component.js
```diff
@@ -138,8 +138,9 @@ class SqlView extends Component {
 
     /* eslint-disable react/no-array-index-key */
     renderCell = (cell, index) => {
-        const isObject = typeof cell === 'object' && cell != null;
-        const formattedValue = isObject ? JSON.stringify(cell, null, 2) : cell;
+        const formattedValue =
+            cell === null ? '' :
+            (typeof cell === 'object' ? JSON.stringify(cell, null, 2) : String(cell));
```
Mohammer5, line 143: This is quite hard to parse. I think this should be outsourced into its own function for better readability:

```js
const formatCellValue = (cell) => {
  if (cell === null) {
    return null
  }

  if (typeof cell === 'object') {
    return JSON.stringify(cell, null, 2)
  }

  return String(cell)
}
```

Mohammer5 (changes requested):

src/SqlView/SqlView.component.js
tomzemp: Logic moved to function

tomzemp: Hi @Mohammer5 : I've moved to logic into a function. If you'd like any other changes, let me know

Mohammer5 (approved):

Mohammer5:
Sorry for the late response.. Somehow this went under the radar..

PR #2159, opened 2022-03-17
test: replace old jest test setup with d2-app-scripts approach

ismay:
I've removed the configuration and dependencies that were used for the old jest test suite. This allows us to use d2-app-scripts to run unit tests with jest.


Mohammer5 (approved):
:muscle: this leaves only 4 tests that we'll have to fix, so that's good.
Do you want to fix them in this PR as well (as this is a draft PR)?

ismay:
> muscle this leaves only 4 tests that we'll have to fix, so that's good. Do you want to fix them in this PR as well (as this is a draft PR)?

I could, but I imagine that's a bit of a potential rabbit hole and I'm already a bit strapped for time. Probably more practical to merge this and address the failing tests separately. What do you think?

Mohammer5:
> Probably more practical to merge this and address the failing tests separately. What do you think?

Exactly my thought

PR #2331, opened 2022-10-06
fix(relationship-type): use default-programStage for event-programs

Birkbjo:
Fixes [DHIS2-13855](https://dhis2.atlassian.net/browse/DHIS2-13855) which prevented saving `relationShipTypes` with constraint-type `Event in program or program stage` to be saved. This is due to an API-change in 2.39.

Removed the ability for tracker-programs to just post a "program". Payload will now always use the `programStage`, and thus we cannot post a program without a `programStage`.

Event-programs will now use the default-programStage instead of `program.id`. 

Backport: https://github.com/dhis2/maintenance-app/pull/2332


tomzemp (approved):
lgtm

PR #2352, opened 2022-10-12
fix(help-links): update help-link mappings

Birkbjo:


Birkbjo:
>
Some links in src/config/inlinehelp-mapping.json have been removed, some added. The first entry that has been removed (edit/categorySection/categoryOptionGroupSet) is a partially existing url (see [here](https://debug.dhis2.org/dev/dhis-web-maintenance/index.html#/edit/categorySection/categoryOptionGroupSet/C31vHZqu0qU)). The docs page that's been linked to doesn't exist, so I suppose it's been removed due to that. If that's the case for all links that have been removed, I guess it's fine for now. But I do wonder if that actually indicates that we lost some docs?

Yeah because that link will match the `"/edit/categorySection": "/metadata.html#create_${objectType}",` link, and get a correct link to https://docs.dhis2.org/en/use/user-guides/dhis-core-version-master/configuring-the-system/metadata.html#create_category_option_group_set

PR #2378, opened 2022-11-10
feat(relationship-type): support attributes and dataElements for display

Birkbjo:
Implements https://dhis2.atlassian.net/browse/DHIS2-13547 in `Program -> Relationship type`
This allows for customization of what attributes (and dataElements for "event in program or program stage") to show in the relationshipWidget in `Capture`.

### DIFF
--- a/src/config/field-overrides/relationshipType.js
+++ b/src/config/field-overrides/relationshipType.js
@@ -186,35 +186,8 @@ class Constraint extends Component {
         }
     }
 
-    componentDidUpdate(prevProps) {
-        const prevEntity = this.getSelectedRelationshipEntity(prevProps);
-        const entity = this.getSelectedRelationshipEntity();
-
-        if (prevEntity !== entity) {
-            console.log('CHANGED ENTITY', entity);
-            const modelTypes = modelTypesForRelationshipEntity[entity];
-
-            // initialize loading state for modelTypes
-            const optionsLoadingState = modelTypes.reduce((acc, modelOpts) => {
-                const modelType = modelOpts.modelType;
-                acc[modelType] = this.shouldRenderModelType(modelType);
-                return acc;
-            }, {});
-            console.log({ optionsLoadingState }, optionsLoadingState);
-            //this.setState({ optionsLoading: optionsLoadingState });
-        }
-    }
-
     isLoading = () => {
-        const optionsLoading = this.isOptionsLoading();
-        console.log(this.state);
-        return this.state.loading || optionsLoading;
-    };
-
-    isOptionsLoading = () => {
-        return Object.values(this.state.optionsLoading).some(
-            loading => loading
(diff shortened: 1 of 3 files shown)


src/forms/form-fields/drop-down-async.js
```diff
@@ -265,6 +270,7 @@ DropDownAsync.propTypes = {
     model: PropTypes.object,
     options: PropTypes.array,
     onOptionsLoaded: PropTypes.func,
+    fieldFilter: PropTypes.string,
```
Birkbjo, line 273: This was added to prevent adding `programStageDataElements`-fields to every place `programs`-dropdown are used. 

This prop should've been added a long time ago, so we wouldn't need the "hacky" `fieldsForReferenceType` above.

src/config/field-overrides/relationshipType.js
```diff
@@ -156,12 +186,44 @@ class Constraint extends Component {
...
+            acc[modelType] = this.shouldRenderModelType(modelType);
+            return acc;
+        }, {});
+        return optionsLoadingState;
+    };
+
+    isLoading = () => {
+        const optionsLoading = this.isOptionsLoading();
+        return this.state.loading || optionsLoading;
+    };
```
HendrikThePendric, line 218:
```suggestion
    isLoading = () => {
        return this.state.loading || this.isOptionsLoading();
    };
```
This is a line less code and will also prevent `this.isOptionsLoading()` from executing.

HendrikThePendric (commented):
I didn't have time for a full review unfortunately, but I think the code is looking very good. Just leaving a comment for now, because this needs a more thorough inspection and also some testing.

src/config/field-overrides/relationshipType.js
```diff
@@ -125,18 +148,25 @@ class Constraint extends Component {
...
             selected,
             loading: true,
             error: false,
+            // loading state for options
+            // these are initialized in componentDidMount and loaded by `DropdownAsync`
+            // shaped like
+            // { program: true, trackedEntityType: true }
+            optionsLoading,
         };
+        this.getOptionsLoadingState();
```
tomzemp, line 162: why is this method being called here again?

Birkbjo:
@tomzemp thanks for the review, and the testing!

> The one thing I noticed when actually using the app though is that the onRemoveItems called from componentDidUpdate seems like it might not be working as intended? For example, when selecting a different program (which lacks the previous attributes), the assignedItems seems not to be updated before submission (that being said, I don't see a problem in the code 🤨):

Seems like we need to use `componentWillReceiveProps`, or else the state-update caused by `onRemoveItems` will not have up to date props when using the `this.props.value` as `prevState` in `handleSelect`.

tomzemp (approved):
looks good! ⛵️

PR #2388, opened 2022-11-18
fix(options): fix attributeValues for options

Birkbjo:
https://dhis2.atlassian.net/browse/DHIS2-14152

Options didn't load the value of the `attributeValue` correctly. This was due to 2 bugs: one bracket was missing in the field-filter, and it used `:owner` (instead of `:all`, like all other requests for `attributeValues` uses) - but I do think it's better to be explicit and include `value` instead of switching to `:all`.


tomzemp (approved):
lgtm 🚀

PR #2430, opened 2023-01-13
fix: remove csv/xml downloads (start)

tomzemp:
See [DHIS2-14482](https://dhis2.atlassian.net/browse/DHIS2-14482)

This removes the XML, CSV download options in maintenance app download. 

@cooper-joe: I've added you as a reviewer in case you want to weigh in, though I understand if this is not worth your time 😅. 

**Before:**
there was a dropdown to choose format:
image

**After:**
there is a note saying that the format is JSON:
image
I've added margin to the initial `The download contains...` text as I thought it looked strange without a margin and the two lines of text.


Birkbjo (approved):

cooper-joe:
Having `Download format: JSON` in it's own line looks a bit like it could or should be user-configurable, which is slightly confusing. I was thinking we could instead change from `The download contains 89 category options` to something like `Downloading 89 category options in JSON format`, but that'll break all translations, so perhaps not worth it.

I think a middle ground would be to change the `Download format: JSON` to use the same grey text color as the `The download contains...`, to show that it's not configurable here. With that change, LGTM.

cooper-joe (approved):

tomzemp:
Updated to use same grey on `Download format: JSON`:
image

PR #2435, opened 2023-01-16
feat(option-set): add support for multi-text valuetype

Birkbjo:
Filtering of optionSet values was added because `MULTI_TEXT` was not supported in the data entry app. This has now been implemented in https://github.com/dhis2/aggregate-data-entry-app/pull/290 . So this PR adds the `MULTI_TEXT` valueType back to the selector. 

`MULTI_TEXT` means that multiple options in the option set can be selected.

`Text with multiple values` has been approved by Lars. Maybe ask @cooper-joe about it as well? 

Should we remove the filtering support all together, or is it OK to just remove the field? Then we have the implementation for a later time. However, with the refactor it's probably not necessary.


tomzemp (approved):
Looks good to me. 

I'd probably opt to remove the blockedProperties filtering now that it's not being used to filter anything, but I don't think it matters much.

_Just confirming:_ is support for MULTI_TEXT DEs/TEAs being implemented for Capture too? So that we don't need to have different logic based on domain type?

Birkbjo:
>Looks good to me.

I'd probably opt to remove the blockedProperties filtering now that it's not being used to filter anything, but I don't think it matters much.

Yeah, sounds good.

> Just confirming: is support for MULTI_TEXT DEs/TEAs being implemented for Capture too? So that we don't need to have different logic based on domain type?

That's a really good question, will investigate.

src/config/field-rules.js
```diff
@@ -1,3 +1,4 @@
+import optionSet from './disabled-on-edit/optionSet.js';
```
tomzemp, line 1: I don't get this import; is it needed here?

src/config/field-rules.js
```diff
@@ -66,6 +67,60 @@ export default new Map([
...
+                            fieldConfig.originalOptions = fieldConfig.props.options;
+                        }
+                        fieldConfig.props.options = options.filter(option => option.value !== 'MULTI_TEXT');
+                    }
+                },
+            }, {
+                type: 'CHANGE_VALUE',
+                field: 'optionSet',
+                setValue: (model, fieldConfig) => {
+                    const isMultiText = model && (model.valueType === 'MULTI_TEXT'
```
tomzemp, line 112: would it just be sufficient to check if the optionSet is `MULTI_TEXT` here? It seems like you cannot ever directly select `MULTI_TEXT`, so it's only being set indirectly when you choose a `MULTI_TEXT` optionSet?

tomzemp (approved):
I approve (again 😀). Looks good; I think the new UX is quite clear 🙌

src/config/field-rules.js
Birkbjo: Hehe, good catch. Probably just something VSCode included when I typed `optionSet`...

src/config/field-rules.js
Birkbjo: You would think so but it's needed when editing a data Element, since the `optionSet`-object isn't fully loaded at first load - it just knows the id - optionSets are loaded afterwards.

PR #2436, opened 2023-01-19
DHIS2-13714 Sorting options in order if all the options are numeric

rithvikreddykoripelli:
https://dhis2.atlassian.net/browse/DHIS2-13714

### DIFF
--- a/src/EditModel/option-set/OptionSorter/optionSorter.js
+++ b/src/EditModel/option-set/OptionSorter/optionSorter.js
@@ -1,12 +1,18 @@
 import { Observable } from 'rxjs';
 
 export default function optionSorter(options, sortProperty, sortOrder = 'ASC') {
-    const sortedOptions = options.sort((left, right) =>
-        (left[sortProperty] || '')
-            .localeCompare(right[sortProperty]));
+    const sortAsNumbers = options.every(option =>
+        isFinite(option[sortProperty])
+    );
+    const sortedOptions = sortAsNumbers
+        ? options.sort(
+              (left, right) => left[sortProperty] - right[sortProperty]
+          )
+        : options.sort((left, right) =>
+              (left[sortProperty] || '').localeCompare(right[sortProperty])
+          );
 
-    return Observable
-        .of(sortOrder === 'ASC'
-            ? sortedOptions
-            : sortedOptions.reverse());
+    return Observable.of(
+        sortOrder === 'ASC' ? sortedOptions : sortedOptions.reverse()
+    );
 }


tomzemp (commented):
This looks great, @rithvikreddykoripelli! Thank you for the contribution 🎉.

I have two questions when I look at this (they're not directly related to your code, they're just about how we want this to work in the maintenance app):

1. Your code is only sorting numerically if all the options are numeric (as you put in the PR title), but I'm not sure from the ticket if we need to be that strict. For example, maybe we would want to use natural sort on a mixed array of strings/number-likes, so that we would sort `[2,1,10,'c','a']` and get `[1,2,10,'a','c']`. With the solution in this PR, since not all options are numeric, we'll get `[1,10,2,'a','c']`. I don't really know which is preferable here (natural sort by default or natural sort only when everything is number-like), so let me check if our UX designer has any thoughts.

2. This fix works to update the sort order, but unfortunately in the maintenance app if I then refresh the page, the sort order looks like it is lost:

sort_options_example


The options are being sorted appropriately, but the request to api/optionSets is paginating in a way that doesn't respect the sort order on the underlying options (hence we have that message: 'List might not represent the accurate order of options due to the availability of pagination'). I just want to ask our backend team if this is something that might be fixable, so that it doesn't confuse users when they refresh the page, and then look at the options in the future. Currently, the sort logic is similar on the backend and frontend, so this problem is less apparent.

rithvikreddykoripelli:
@tomzemp , thanks for taking time to review the PR. 

Sure, I will be waiting to hear from you.

tomzemp:
Thanks for your patience @rithvikreddykoripelli

I've spoken about this internally and have some updates:

1. We think it would be better to use natural sort generally rather than using this only when all options are numeric. This should also be more efficient as we won't have to check all the options before defining the sort technique. Would you like to update your implementation to do this? I think the easiest approach would be just to pass `{numeric: true}` as an option to the localeCompare method.
2.  We should actually be able to fix this behaviour (the options not appearing sorted on refresh) by changing our api call. I've opened a new ticket for this [DHIS2-14602](https://dhis2.atlassian.net/browse/DHIS2-14602), which we can implement independently of this ticket/PR.

rithvikreddykoripelli:
Hi @tomzemp , 
Yes, it makes more sense to implement it that way. Sure, I will make the changes accordingly and create a new PR.

tomzemp:
Closed and replaced by: https://github.com/dhis2/maintenance-app/pull/2441

PR #2442, opened 2023-02-08
chore: fix app for ARM64 architecture

HendrikThePendric:
I've bought a new Mac and found I couldn't boot this old app. Underlying dependencies not being available for my new architecture was the problem. In the end updating the Sass library was required, which then resulted in some Sass compilation errors, which were easy enough to fix. The changes are pretty minimal, but devs on a silicon based Mac can work on this app with these fixes applied.


package.json
```diff
@@ -64,8 +64,8 @@
     "loglevel": "^1.4.0",
     "moment": "^2.16.0",
     "node-fetch": "^1.6.3",
-    "node-pre-gyp": "^0.6.30",
-    "node-sass": "^4.13.1",
+    "node-pre-gyp": "^0.17.0",
+    "node-sass": "npm:sass@^1.58.0",
```
HendrikThePendric, line 68: `node-sass` was the cause of the problem. But that package was deprecated and replaced by `sass`, which was supposed to be a drop in replacement. However, `node-pre-gyp` was still referring to `node-sass`. So creating an alias turned out to be possible and a good solution imo.

scss/EditModel/EditDataEntryForm.scss
```diff
@@ -133,6 +134,3 @@
   }
 }
 
-.expand {
-  @extend .item.active;
-}
```
HendrikThePendric, line 138: This syntax was no longer considered valid, so I changed it to the above, which is functionally equivalent.

Mohammer5: This is only functionally when `.expand` is inside `.item`, which does not have to be the case in the old code.
If you know that this will always be the case, then the change is OK, otherwise it has to to sth like this:

```scss
.item.active, .expand {
  a {
    cursor: pointer;
    &:hover {
      text-decoration: underline;
    }
  }
}

.item.inactive {
  a {
    color: inherit;
    opacity: 0.5;
    cursor: default;
  }
}
```

scss/EditModel/EditDataEntryFormProgramStage.scss
```diff
@@ -110,7 +110,8 @@
   }
 
   .item {
-    &.active {
+    &.active,
+    &.expand {
```
Mohammer5, line 114: Same as above

Mohammer5 (changes requested):

Mohammer5 (approved):

PR #2444, opened 2023-02-24
fix(programNotifications): move webhook url to recipient step

Birkbjo:
Previously it was quite awkward to use `WEB_HOOK` recipient. Since the web-hook URL is re-using the `messageTemplate` in the model - you would have to go to step 3, choose `Web hook` as recipient, then go back to first step to set the URL.

This PRs adds a custom override for `notificationRecipient` that also renders the `messageTemplate`-field when `WEB_HOOK` is selected (with a custom label for Web Hook URL) so this field is "moved" to the third step.


src/config/field-overrides/programNotificationTemplate.js
```diff
@@ -129,6 +132,38 @@ const DataElementDropDown = compose(
...
+    )
+)(({ onUpdate, d2, ...props }) => {
+    const label = d2.i18n.getTranslation('web_hook_url');
+    if (props.value === 'WEB_HOOK') {
+        return (
+            <div>
+                <Dropdown {...props} />
+                <TextField
+                    label={label}
+                    multiLine
```
amcgee, line 149: minor: should this be multiline...?

amcgee:
Code looks good to me, one minor comment - I'm not an expert on this app structure and haven't tested myself but the change looks simple enough, nice work @Birkbjo !

src/config/field-overrides/programNotificationTemplate.js
Birkbjo: Good catch. Most of this was copy pasted from the subject template component. Thanks!

HendrikThePendric (approved):
Great work, thanks @Birkbjo

tomzemp (approved):
I approve, but I think the logic of how this is handled on the backend is super weird and then contributes to strange UX here. If we're storing the URL in the messageTemplate, it seems odd to let people enter something that is not a URL and then use that value once you've selected the WEB_HOOK option. Personally, I feel like if you select WEB_HOOK, it would probably better to just clear out what you had previously entered in messageTemplate field? Like who is going to know that the messageTemplate is going to be reused as destination URL when they select WEB_HOOK?

Birkbjo:
> Personally, I feel like if you select WEB_HOOK, it would probably better to just clear out what you had previously entered in messageTemplate field? Like who is going to know that the messageTemplate is going to be reused as destination URL when they select WEB_HOOK?

Yeah, that's a valid point. But could also be a bit annoying if a user had already filled in the message template, and accidentally selected `Web hook`, having that template value cleared. Probably rare, but still.

I do think the `Who to send to`-step should come first, and could clear up some of these issues. Like how every messaging/email client works... But that is probably too drastic of a change - but something we could keep in mind in the refactor.

PR #2445, opened 2023-02-27
fix: sort options by sort order [DHIS2-14602]

tomzemp:
This PR updates option set ordering in the maintenance app to sort options by sort order (and removes the existing messaging about option being potentially displayed in the incorrect order).

(note this helps facilitate the PR for [DHIS2-13714](https://dhis2.atlassian.net/browse/DHIS2-13714), so the we can use natural sort and then display the options back to the user in the correct order)

**Before**

image

**After**
image

[DHIS2-13714]: https://dhis2.atlassian.net/browse/DHIS2-13714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/EditModel/option-set/OptionManagement.component.js
+++ b/src/EditModel/option-set/OptionManagement.component.js
@@ -107,21 +107,6 @@ class OptionManagement extends Component {
         translate: true,
     });
 
-    displayInCorrectOrderWarning() {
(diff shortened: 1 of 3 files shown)


Birkbjo (approved):
Nice, thanks!

PR #2447, opened 2023-02-28
fix: add missing translations and clean custom columns

Birkbjo:
A lot of properties in "Manage Columns" were not translated. 

Gone through every object (I think), and added missing translations. Also added some properties to the ignore-list when I think they were unnecessary. 

Fixes https://dhis2.atlassian.net/browse/DHIS2-14042


tomzemp (approved):
Good job finding all the missing translations 🙌!

I'm not 100% sure on the column removes (like `orgUnitLevelName` is still on the schema for dataApprovalLevels, so I don't know if it's used or not 🤷‍♂️); I take your word for it, but just point this out in case you wanted a review from someone with more certainty.

Birkbjo:
> I'm not 100% sure on the column removes (like `orgUnitLevelName` is still on the schema for dataApprovalLevels, so I don't know if it's used or not 🤷‍♂️); I take your word for it, but just point this out in case you wanted a review from someone with more certainty.

Should probably have made the comment clearer - it always returns an empty string so I didn't see a reason to keep it around.

tomzemp:
Got you. Makes sense

PR #2448, opened 2023-03-01
feat: introduce org-unit-field override component

HendrikThePendric:
Fixes [DHIS2-13374](https://dhis2.atlassian.net/browse/DHIS2-13374)

@jimgrace and @Bekkalizer this is the addition of an `orgUnitField` to the Program Indicator form in the maintenance app. It would be good if you could also test this once it's gone to master.

[DHIS2-13374]: https://dhis2.atlassian.net/browse/DHIS2-13374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/config/field-overrides/program-indicator/OrgUnitField.js
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    componentDidUpdate() {
+        if (
+            this.hasRelevantPropChanged() &&
+            this.shouldFetchDataElementsForProgram()
+        ) {
+            this.fetchDataElements();
+        }
+
```
Birkbjo, line 88: Currently there's a bug causing the old `value` to linger if you change the program. And in some cases this value shouldn't be a valid selection if program is changed.

So I think something like this is needed:
```js
      if(this.hasProgramChanged()) {
            this.onChange({target: { value: this.getDefaultValue()}})
        }
```

src/config/field-overrides/program-indicator/OrgUnitField.js
```diff
@@ -0,0 +1,323 @@
...
+        });
+
+        try {
+            const api = this.d2.Api.getApi();
+            const url = `programs/${this.props.model.program.id}`;
+            const data = {
+                fields:
+                    'programStages[id,programStageDataElements[dataElement[id,displayName,valueType]]',
+            };
+            const allDataElements = await api.get(url, data);
```
Birkbjo, line 127: The more "d2" or "maintenance" way to do this would be something like

```js
const allDataElements = await this.d2.models.program.get(this.props.model.program.id, { fields: data.fields})
```

Shouldn't really matter though, but then you don't have to concat the url yourelf.

Birkbjo (changes requested):
Overall looks very good! Very clean as well - would clean something like this would be more complex, so good job!

Most comments are minor, but I did encounter a few bugs. Notably the field rendering in all steps, and not resetting the value of the field when program is changed.

src/config/field-overrides/program-indicator/OrgUnitField.js
```diff
@@ -0,0 +1,323 @@
...
+        }
+    }
+
+    render() {
+        if (!this.hasRequiredParams()) {
+            return null;
+        }
+
+        return (
+            <DropDown
```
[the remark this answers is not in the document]
HendrikThePendric: Sorry, I never even checked the other steps 😊 
Thanks for spotting this

src/config/field-overrides/program-indicator/OrgUnitField.js
```diff
@@ -0,0 +1,323 @@
...
+            this.setState({
+                dataElements: {
+                    ...this.state.dataElements,
+                    [this.props.model.program.id]: orgUnitFieldDataElements,
+                },
+                loading: false,
+                error: null,
+            });
+        } catch (error) {
+            this.setState({
```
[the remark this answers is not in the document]
HendrikThePendric: Yes, this is a pet peeve of mine too, so omitting is very embarrassing 🙄 .

src/config/field-overrides/program-indicator/OrgUnitField.js
```diff
@@ -0,0 +1,323 @@
...
+                    'programStages[id,programStageDataElements[dataElement[id,displayName,valueType]]',
+            };
+            const allDataElements = await api.get(url, data);
+            const orgUnitFieldDataElements = allDataElements.programStages
+                .flatMap(({ programStageDataElements }) =>
+                    programStageDataElements.map(
+                        ({ dataElement }) => dataElement
+                    )
+                )
+                .filter(({ valueType }) => valueType === ORG_UNIT_VALUE_TYPE);
```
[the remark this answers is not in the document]
HendrikThePendric: In the end I decided to combine this comment with your other one here https://github.com/dhis2/maintenance-app/pull/2448#discussion_r1121988211 and cleaned the code up a bit here: https://github.com/dhis2/maintenance-app/pull/2448/commits/d01bf206a1c63ae88a1e0ae973086e650984704b

src/config/field-overrides/program-indicator/OrgUnitField.js
HendrikThePendric: After discussing some details and edge cases with Birk on Slack. I've implemented the following changes:
* I am holding on to the selected value, unless it's not among the new dropdown-options, in which case I reset to default (so basically option 2 above).
* Since all of this resetting can be a bit confusing to a user who may or may not be aware of what is going on behind the scenes, I thought it would be beneficial to add a `(default)` suffix in the labels of the default options, so that is clear for users that we're resetting to the default. 
* This also means that we now have two "flavours" of enrolment options: 
    * `Enrollment organisation unit (default)` which is the default when `analyticsType === ENROLLMENT`
    * `Enrollment organisation unit` which is an option the user may select explicitly when `analyticsType === EVENT`

src/config/field-overrides/program-indicator/OrgUnitField.js
```diff
@@ -87,18 +93,40 @@ export default class OrgUnitField extends Component {
...
+            const isCurrentValueInOptions = options.some(
+                ({ value }) => value === this.props.value
+            );
+
+            // Reset selection to default if selected option is no longer valid
+            if (!isCurrentValueInOptions) {
+                this.props.onChange({ target: { value: undefined } });
+            }
+
+            this.setState({ options });
```
Birkbjo, line 127: Is this used only for the component do update? If so I think a comment explaining that is warranted.

src/config/field-overrides/program-indicator/OrgUnitField.js
```diff
@@ -175,7 +203,7 @@ export default class OrgUnitField extends Component {
         return false;
     }
 
-    shouldFetchDataElementsForProgram() {
+    requiresDataElementsForProgram() {
```
Birkbjo, line 206: The comments in this function doesn't really match the name of the function anymore. I actually think the previous name was a bit clearer?
You could just keep the `requiresDataElementsForProgram` as name of the variable of the result of this.

src/config/field-overrides/program-indicator/OrgUnitField.js
```diff
@@ -87,18 +93,40 @@ export default class OrgUnitField extends Component {
...
+    componentDidUpdate(_, prevState) {
+        const isDoneLoading =
+            prevState.loading === true && this.state.loading === false;
+        const hasRelevantPropChanged = this.hasRelevantPropChanged();
+        const requiresDataElementsForProgram = this.requiresDataElementsForProgram();
+        const shouldFetch =
+            hasRelevantPropChanged && requiresDataElementsForProgram;
+        const shouldSetOptions =
+            (isDoneLoading || hasRelevantPropChanged) &&
+            !requiresDataElementsForProgram;
```
Birkbjo, line 110: It's not immediately obvious why `requiresDataElementsForProgram` decides if we should set the options or not. 
And after some more thinking, I'm not sure if we actually need that check? We could just `setOptions` regardless, right? Yeah there's a few more times it would run, but does that matter?

Birkbjo (approved):
Looks very good! Just some minor comments.

src/config/field-overrides/program-indicator/OrgUnitField.js
HendrikThePendric: You may be correct... But I'm not actually sure.... I think that we probably do need to include `requiresDataElementsForProgram ` in this check.... We need to consider these two cases:
1. The relevant props have changed and no data elements need to be fetched (this could either be because the combination of program+analyticsType doesn't need data elements, or because the relevant data elements had been fetched before and are still available). In this case we're ready to update the options straight away.
2. Or the relevant props changed and for this combination of program+analyticsType we do need to fetch data elements. In this case we need to wait until the data elements are returned from the server, which is an "event" we can identify be the `loading` state transitioning from `true` to `false`

So if I were to write this condition to strictly reflect these two scenarios, I would end up with a slightly different comparison operator, but still including `requiresDataElementsForProgram`:
```js
const shouldSetOptions = (hasRelevantPropChanged && !requiresDataElementsForProgram) || 
     isDoneLoading;
```

But I took a slightly different approach, which I think also makes sense. In the check I try to
1. Did an "event" (aka props or state transition) take place which requires the options to be recomputed?
    ```js
    (isDoneLoading || hasRelevantPropChanged)
    ```
2. And do we have enough information available to actually compute the new `options`?
    ```js
    !requiresDataElementsForProgram
    ```

I actually still think this is a sound approach.

So I'd probably argue that I should keep this check as-is. But change the variable and or method name, because that is what seems to be causing confusion here....

Birkbjo: > I actually still think this is a sound approach.

> So I'd probably argue that I should keep this check as-is. But change the variable and or method name, because that is what seems to be causing confusion here....

Sounds good to me, it just wasn't immediatly obvious that that `!requiresDataElementsForProgram` is used for ```do we have enough information available to actually compute the new options?```.

HendrikThePendric: A compelling argument for keeping the current approach is an edge case we could encounter when moving to the alternative approach I mention above: 
Suppose the user's internet connection is very slow and he/she change the program in the form multiple times in quick succession. If both programs require data elements `loading` will transition from `true` to `false` twice and if we try to set the options after the first transition it'll try producing the options for the second program, but its data elements are not available yet.

PR #2468, opened 2023-04-25
fix(programindicators): default period boundary

Birkbjo:
Fixes https://dhis2.atlassian.net/browse/DHIS2-15205

I'm a bit confused by the previous logic for setting the default boundaries, and it didn't seem to comply to the specs, since it did not correctly reset the boundaries when `analyticsType` changed. It would only reset the boundaries if they were equal to some default - but that is the origin of the bug described in the ticket above as well; it would prevent setting the boundaries that happened to be equal to the other types default (eg. type `Enrollment`, and both boundary targets `Event date` .

It might've been a workaround due it being hard to know if only the `analyticsTápe` changed without a variable for it - so I included that in a small hack.

### DIFF
--- a/src/config/field-config/field-defaults.js
+++ b/src/config/field-config/field-defaults.js
@@ -1,5 +1,3 @@
-import { isEqual } from 'lodash/fp';
-
 import {
     BEFORE_END_OF_REPORTING_PERIOD,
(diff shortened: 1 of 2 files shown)


KaiVandivier (approved):
Tested locally and seems to be fixed 🙂

PR #2549, opened 2023-06-16
fix: update email validator [DHIS2-15417]

tomzemp:
See [DHIS2-15417](https://dhis2.atlassian.net/jira/software/c/projects/DHIS2/issues/DHIS2-15417) including first comment.

This updates the email validator as it seems to be missing a lot of invalid email addresses which then allows user to save (and subsequently get a message saying the update was rejected because of an invalid email).

[DHIS2-15417]: https://dhis2.atlassian.net/browse/DHIS2-15417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/forms/fields.js
```diff
@@ -43,6 +42,17 @@ function isIntegerValidator(value) {
 }
 isIntegerValidator.message = 'number_should_not_have_decimals';
 
+function isEmpty(value) {
+    return typeof value === 'undefined' || value === null || value === ''
+}
+
+const EMAIL_ADDRESS_PATTERN =
```
Birkbjo, line 49: Do we have a source for this? Is it the same as the backend pattern?

Birkbjo (approved):

src/forms/fields.js
tomzemp: @Birkbjo : I put a bunch of the notes for this ticket in the jira:

> I looked into this and we are using [EmailValidator](https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/EmailValidator.html) for backend check, which unfortunately does not seem to provide much information about what it is actually doing, so it’s difficult to make the check behave exactly the same on both the backend and front end.

> That said, our client-side validation check [here](https://github.com/dhis2/d2-ui/blob/fdb5481eb7fb3777eed41ad85c72c34477694340/packages/forms/src/Validators.js#L8C30-L9) seems not to be working appropriately.  

> We updated the RegEx check in our more up-to-date UI library ([ui/collections/forms/src/validators/email.js at master · dhis2/ui](https://github.com/dhis2/ui/blob/master/collections/forms/src/validators/email.js#L28) ), and moving to this seems like it will fix a bunch of the issues where the maintenance app is clearly failing to flag invalid email addresses.

As I can't find the code for the backend logic, I don't think we can replicate that. I have just copy/pasted (with light edits) the existing ui library's front-end validator, since I didn't think we'd want to introduce a new dependency in the legacy maintenance app? But I can try to import it, if you disagree. Or if you don't think that's necessary, I can add a comment about where it's from.

tomzemp: Added comments

PR #2616, opened 2023-06-28
fix: update email validator [DHIS2-15417] (#2549)

tomzemp:
Backport of https://github.com/dhis2/maintenance-app/pull/2549

### DIFF
--- a/src/forms/fields.js
+++ b/src/forms/fields.js
@@ -2,7 +2,6 @@ import {
     isRequired,
     isUrl,
     isNumber as isNumberValidator,
-    isEmail,
 } from 'd2-ui/lib/forms/Validators';
 import isString from 'd2-utilizr/lib/isString';
 import { getOr, isNumber } from 'lodash/fp';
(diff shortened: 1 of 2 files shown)


src/forms/fields.js
```diff
@@ -2,7 +2,6 @@ import {
     isRequired,
     isUrl,
     isNumber as isNumberValidator,
-    isEmail,
 } from 'd2-ui/lib/forms/Validators';
 import isString from 'd2-utilizr/lib/isString';
 import { getOr, isNumber } from 'lodash/fp';
```
Birkbjo, line 7:
```suggestion
import { getOr, isNumber, isEmpty } from 'lodash/fp';
```

src/forms/fields.js
```diff
@@ -43,6 +42,19 @@ function isIntegerValidator(value) {
 }
 isIntegerValidator.message = 'number_should_not_have_decimals';
 
+function isEmpty(value) {
+    return typeof value === 'undefined' || value === null || value === ''
+}
```
Birkbjo, line 47: Should be able to use `isEmpty` `lodash/fp`, as suggestion above?

Birkbjo (approved):
Just a minor comment.

src/forms/fields.js
tomzemp: I'm going to merge as is because this code is already in master branch. I took the code for the validator from dhis2/ui (where there's no lodash dependency and isEmpty is a helper function: https://github.com/dhis2/ui/blob/master/collections/forms/src/validators/helpers/index.js#L2).

PR #2624, opened 2023-06-29
fix(dataelement): remove multi-text optionset restriction for tracker

Birkbjo:
The restriction to prevent selection of an `optionSet` with `MULTI_TEXT` when `aggregationType = TRACKER` was added because multi-selects where not implemented in tracker. 
This PR https://github.com/dhis2/capture-app/pull/3231 implements that, so it's time to remove the restriction.


simonadomnisoru (approved):
LGTM

PR #2688, opened 2023-08-16
feat(optionset) option search

Birkbjo:
https://dhis2.atlassian.net/browse/DHIS2-13440

### DIFF
--- a/src/EditModel/option-set/OptionManagement.component.js
+++ b/src/EditModel/option-set/OptionManagement.component.js
@@ -87,7 +87,7 @@ class OptionManagement extends Component {
 
     componentWillReceiveProps(newProps) {
         if (this.props.model !== newProps.model) {
-            actions.getOptionsFor(newProps.model, undefined);
+            actions.getOptionsFor(newProps.model, this.state.filter);
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/EditModel/option-set/OptionManagement.component.js
```diff
@@ -90,6 +97,11 @@ class OptionManagement extends Component {
         }
     }
 
+    handleFilter = ({ target: { value } }) => {
+        this.setState({ filter: value })
+        actions.getOptionsFor(this.props.model, value);
```
Birkbjo, line 102: Note that this is of course debounced. The easy debouncing of observables is something I will actually miss about this paradigm.

Birkbjo:
> the filter persists in the component state when you add a new option, but it's not applied to the options. I guess passing the filter in the componentWillReceiveProps would "solve" this issue by keeping the filter applied, though maybe ideal UX would be to clear the filter at this point to make sure that people see whatever they just added

I think I will just pass the filter in `componentWillReceiveProps`. I think it's worse to clear the filter, because that filter probably won't end up in the first page anyway. 

> it might be useful to trim the filter input (e.g. for communicating that you've applied the filter); if you just add spaces as the filter, it doesn't seem to actually affect the options returned, but it does disable the sort buttons, and if you click away from the filter, it doesn't look like anything is there

I did initially do this. But it causes some issues. If you trim it directly, you can't enter spaces. You don't actually want to make these changes to the filter-state, because it's weird to not being able to add as many spaces as you want?
 You could check `value.trim().length !== 0`, though. But that cases issues when clearing the filter.  

So I will just add a check for the disabled, and if `filter.trim().length === 0`, we don't disable the buttons.

PR #2802, opened 2024-02-20
feat(DHIS2-16132): add ability to transpose/pivot a section form

kabaros:
This PR adds configuration option to the maintenance app to configure a section to be transposed, and grouped by data elements in the data entry app. Implements [DHIS2-16132 ](https://dhis2.atlassian.net/browse/DHIS2-16132).

[pivot-options-2.webm](https://github.com/dhis2/maintenance-app/assets/1014725/89e82ef0-2ac9-45d6-8c2a-1b816ddcc63d)

## Background

This is [an RFC](https://dhis2.atlassian.net/wiki/spaces/SOFTWARE/pages/121995265/Form+configuration+RFC) that describes the approach and the priorities for form configuration options. This is based on a [thorough investigation](https://docs.google.com/presentation/d/1pSeFoF91HDYNeL88GghKLvj3wpYf0qUCdhkKXklB2MY/edit#slide=id.g2462c2dd6c9_0_53) by the functional design team for custom form use cases in real-life implementations. Based on that investigation, the ability to transpose forms and adding free-text descriptions were one of the main reasons people choose to go the custom forms route so we're tackling these first.

## Related PRs:

- [Maintenance app](https://github.com/dhis2/maintenance-app/pull/2802)
- [Data entry app (beta)](https://github.com/dhis2/aggregate-data-entry-app/pull/367)
- [Backend](https://github.com/dhis2/dhis2-core/pull/16562)


cooper-joe (approved):
LGTM, one minor comment.

src/EditModel/SectionDialog.component.js
```diff
@@ -216,40 +306,136 @@ class SectionDialog extends React.Component {
...
+                                value={value}
+                                label={text}
+                                style={{ margin: '10px' }}
+                            />
+                        );
+                    })}
+                </RadioButtonGroup>
+                {isMoveCategoryMode && (
+                    <SelectField
+                        style={{ marginLeft: '50px' }}
```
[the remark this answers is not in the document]
kabaros: updated

image

image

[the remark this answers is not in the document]
cooper-joe: Thanks, that looks good.

src/EditModel/SectionDialog.component.js
```diff
@@ -216,40 +306,136 @@ class SectionDialog extends React.Component {
...
                     action: this.getTranslation('ok'),
                 });
             });
-    }
+    };
+
+    handleChoosePivotMode = (e, pivotMode) => {
+        const pivotedCategory =
+            pivotMode === 'move_categories'
+                ? this.state.categories[0].id
```
Birkbjo, line 350: I think we should be a bit defensive here - what if there's no categories in the DB, or if the request fails?

src/EditModel/SectionDialog.component.js
```diff
@@ -26,176 +29,263 @@ class SectionDialog extends React.Component {
...
+                this.forceUpdate();
+            })
+        );
+
+        const d2 = this.context.d2;
+        const categoryCombos = this.props.categoryCombos
+            .map(coc => coc.value)
+            .join(',');
+
+        return d2.models.categoryCombos
```
Birkbjo, line 62: You should be able to get this data directly from `categories`-endpoint instead of having to gather the categories from the catcombos yourself.

Something like this: `https://debug.dhis2.org/dev/api/categories.json?fields=id,displayName,categoryCombos[id]&filter=categoryCombos.id:in:[DJXmyhnquyI,bjDvmb4bfuf]`


```js

return d2.models.categories.list({ filter: `categoryCombos.id:in:[${categoryCombos}]`, fields: ['id,displayName] })

```

```

Birkbjo: I also think we should filter out the default category, right? It doesn't make much sense to select that, right? The default category can only be by itself, and thus you should just pivot the form instead of moving it.

Birkbjo (changes requested):

Birkbjo:
Functionality wise looks really good. One minor comment, and I think the data-fetching could be made easier - but I might be missing something!

src/EditModel/SectionDialog.component.js
kabaros: I updated the two comments (about fetching and being defensive) - thanks for these!

Regarding the default category, it's a bit fiddly because sections can have data elements with different category combos, so it's not necessarily clear whether it's the only option or not for a specific section. I need to figure that usecase better, and I will have a followup PR if that's ok - for now, they should be able to "preview" the change and see if something is wrong with the configuration until we tighten the option per section better.

Birkbjo (approved):
LGTM! 🥳

PR #2812, opened 2024-03-06
feat(DHIS2-17002): configure text to display before and after section

kabaros:
This PR adds the ability to configure custom HTML text before and after a section. This implements [DHIS2-17002
](https://dhis2.atlassian.net/browse/DHIS2-17002)

Related data-entry PR: https://github.com/dhis2/aggregate-data-entry-app/pull/373

[text_before_after_section.webm](https://github.com/dhis2/maintenance-app/assets/1014725/7130f3c9-959d-40d0-a175-3522377c33ff)


src/EditModel/SectionDialog.component.js
```diff
@@ -359,6 +359,16 @@ class SectionDialog extends React.Component {
         });
     };
 
+    handleSectionTextUpdate = (fieldName) => (ev, text) => {
+        console.log('>>>>>>', fieldName, text)
```
Birkbjo, line 363: Remove?

Birkbjo (approved):
Minor comment, but looks good to me!

kabaros:
@Birkbjo I made [this change](https://github.com/dhis2/maintenance-app/compare/abfcdce4509f2d0947dd4d11158dde777bed8188..6771932a404ace584d88f580acba3d532455491f) based on Michael [comment](https://github.com/dhis2/aggregate-data-entry-app/pull/373#issuecomment-1989486919) about cleaning the HTML on entry - let me know if it looks ok

src/EditModel/SectionDialog.component.js
```diff
@@ -359,6 +361,17 @@ class SectionDialog extends React.Component {
         });
     };
 
+    handleSectionTextUpdate = (fieldName) => (ev, text) => {
+        const cleanHtml = DOMPurify.sanitize(text);
```
Birkbjo, line 365: Should we call this in `componentWillReceiveProps` as well? So that the initial-value is sanitized as well? 

Of course this should ideally be done server-side, but this could help if someone opened a section that has been changed through API or some other means (and don't change the text).

Maybe we should extract it to it's own method?

```js

sanitizeDisplayOptions(displayOptions) //could be static or outside, idk

// in this code:

const newDisplayOption = sanitizeDisplayOptions({ displayOptions: { ...this.state.displayOptions, [fieldName]: cleanHtml }})

// componentWillReceiveProps

  displayOptions: props.sectionModel.displayOptions
                        ? this.sanitizeDisplayOptions(JSON.parse(props.sectionModel.displayOptions))
                        : {},

```

If you think it's unnecessary to always sanitize both keys, I'm fine with also sanitizing them directly.

```js
sanitizeText(text)

const newDisplayOptions = {
...this.state.displayOptions,
[fieldName]: sanitize(text)
}

// would make componentWillReceiveProps handling a bit more convoluted though

```

Just a bit more convenient to pass the entire thing.

Birkbjo: Disregard my last comment. I think it would be better to just run sanitization in `saveSection`, right?

src/EditModel/SectionDialog.component.js
```diff
@@ -359,6 +374,15 @@ class SectionDialog extends React.Component {
         });
     };
 
+    handleSectionTextUpdate = (fieldName) => (ev, text) => {
+        this.setState({
```
Birkbjo, line 378: To be pedantic this should use the callback-style for prevValue, right?
Shouldn't really matter, though...

Birkbjo (approved):

PR #2818, opened 2024-03-18
feat(icons): support uploading icons

Birkbjo:
[14276](https://dhis2.atlassian.net/browse/DHIS2-14276)

Added support of upload of icons. The `/icon` api is now paginated, and thus I had to implement this, and do server-side search.

- The app has concepts of "variance", eg. "positive", "negative", "outline". However there's no concept of this in the API. Previously this was sorted and grouped on the client, however we can't do that anymore due to pagination. 
     - So I changed these tabs to basically just be a "prefilled"-filter. It's not very elegant, since you can't combine this with a search... And you will get some false-positives (eg. some icons have "negative" in the keyword, which refers to eg. a test-negative, not the image being of "negative"-variance.


src/forms/form-fields/helpers/IconPickerDialog/IconPickerCustomTab.js
```diff
@@ -0,0 +1,210 @@
...
+            iconMetadata: {
+                ...prevState.iconMetadata,
+                [name]: value,
+            },
+        }));
+    };
+
+    handleIconKeywordsChange = event => {
+        const keywords = event.target.value
+            .split(',')
```
kabaros, line 63: maybe we can be lenient here and split by new lines too - this seems like my default go-to (and the hint text is not very obvious)
```suggestion
          .split(/[,(\r?\n)]+/)
```

src/forms/form-fields/helpers/IconPickerDialog/IconPickerCustomTab.js
```diff
@@ -0,0 +1,210 @@
...
+        // since this is async, we can't do it during render
+        const iconFileBase64 = await fileToBase64(file);
+
+        this.setState(prevState => ({
+            iconFile: file,
+            iconFileBase64,
+            iconMetadata: {
+                ...prevState.iconMetadata,
+                // remove file extension and replace hyphens with underscores
+                key: file.name.replace(/\..*$/, '').replaceAll('-', '_'),
```
kabaros, line 103: maybe remove spaces too since the API doesn't like them, and files often have them

src/forms/form-fields/helpers/IconPickerDialog/IconPickerCustomTab.js
```diff
@@ -0,0 +1,210 @@
...
+                                {this.context.d2.i18n.getTranslation(
+                                    'choose_file_to_upload'
+                                )}
+                            </span>
+                        </Button>
+                        <input
+                            type="file"
+                            onChange={this.handleFileChange}
+                            style={styles.iconFileInput}
+                            ref="iconFileInput"
```
kabaros, line 157: do we only accept pngs? (this confused me because the fallback icon is svg)

src/forms/form-fields/helpers/IconPickerDialog/IconPickerCustomTab.js
```diff
@@ -0,0 +1,210 @@
...
+                    this.props.onIconUpload(iconData.key);
+                }
+                this.setState({ uploading: false, uploadError: null });
+            })
+            .catch(e => this.setState({ uploading: false, uploadError: e }));
+    };
+
+    handleFileChange = async event => {
+        const file = event.target.files[0];
+        if (!file) {
```
kabaros, line 89: it would probably be good to limit `file.size` to something reasonable (in the future)

kabaros (approved):
Some suggestions here but it looks good to me - tested locally as well.

src/forms/form-fields/helpers/IconPickerDialog/IconPickerCustomTab.js
Birkbjo: Added returns. Not sure about the `()` and `?`, though?

src/forms/form-fields/helpers/IconPickerDialog/IconPickerCustomTab.js
Birkbjo: Added `\s` and `()` to be replaced.

src/forms/form-fields/helpers/IconPickerDialog/IconPickerCustomTab.js
Birkbjo: There's server side validation for this: 
image

Want me to add that to client as well? I do think this could be changed in server config, so maybe enough to depend on the server for this?

src/forms/form-fields/helpers/IconPickerDialog/IconPickerCustomTab.js
kabaros: the `?` as `\r` is present on windows, but not on linux/mac .. but I think your code would work anyhow since there is a trim after that will get rid of either \r or \n (or both) .. so you could split by just `\n`

Birkbjo: As far as I understand the `?` and `()` would be replaced literally. Eg. split on `?` and parenthesis?

Birkbjo:
```js
`is this
a string?
Yes?Iam,am(split)`.split(/[,(\r?\n)]+/)
```

Results in:
```js
['is this', 'a string', 'Yes', 'Iam', 'am', 'split', '']
```

PR #2828, opened 2024-04-10
fix(program): migrate from legacy sharing API

Birkbjo:
https://dhis2.atlassian.net/browse/DHIS2-17210
Fixes https://dhis2.atlassian.net/browse/BETA-153
Enables the app to handle the new "sharing"-structure. Note that this will still use the "old" sharing-dialog, that does not use the JSON-patch method of updating sharing. 
The most important to fix is the references in models to the old structure - import APIs will still work. 

Sorry for the prettier styling fixes...

 * Implemented transformers to convert between old and new structures
     * This is mainly used because the sharing dialog expects the old format, and we do not have control over that code in this app.
 * Updated the "access"-step in Programs to work on `sharing` object instead of old properties. 
 * In `List`: maps `sharing.public` to `publicAccess`
     * Not feasible to update references here, because Datatable and its rendering of `PublicAccess` are not controlled by the app. It's also not trivial to support "nested" paths for controlling that through `valueRenderer`

### DIFF
--- a/src/EditModel/event-program/program-access/ProgramStagesAccess.js
+++ b/src/EditModel/event-program/program-access/ProgramStagesAccess.js
@@ -18,6 +18,10 @@ import {
     generateSharingDescription,
 } from './utils';
 import { yellow800 } from 'material-ui/styles/colors';
+import {
+    transformLegacySharingToSharingObject,
+    transformSharingObjectToLegacy,
+} from '../../sharing';
 
 const styles = {
     container: {
@@ -58,8 +62,6 @@ const styles = {
     },
 };
 
-const sharingFields = ['publicAccess', 'userAccesses', 'userGroupAccesses'];
-
 class ProgramStagesAccess extends React.Component {
     constructor(props) {
         super(props);
@@ -67,15 +69,13 @@ class ProgramStagesAccess extends React.Component {
         // Get (supersets of) sharing structures
         const programSharing = this.props.model.dataValues;
         const stagesSharing = Array.from(
-            this.props.model.dataValues.programStages.valuesContainerMap,
+            this.props.model.dataValues.programStages.valuesContainerMap
         ).map(stage => stage[1]);
 
         // Pre-select stages with similar sharing settings as program
         const selectedStages = [];
-        const stagesWithSimilarAccessAsProgram = [];
         stagesSharing.forEach(stage => {
             if (areSharingPropertiesSimilar(programSharing, stage)) {
-                stagesWithSimilarAccessAsProgram.push(stage.id);
                 selectedStages.push(stage.id);
             }
         });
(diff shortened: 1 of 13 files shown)


src/EditModel/event-program/program-access/ProgramStagesAccess.js
```diff
@@ -157,114 +154,101 @@ class ProgramStagesAccess extends React.Component {
...
-
-        this.setState(
-            {
-                stagesSharing: this.state.stagesSharing.map(propagateIfSelected),
-            },
-            this.updateInternalState,
-        );
-    };
-
-    updateInternalState = () => {
```
Birkbjo, line 237: This is a typical example where this should just be calculated instead of put in state. Only used in render.

src/EditModel/sharing.js
```diff
@@ -0,0 +1,103 @@
...
+
+    return {
+        public: publicAccess,
+        external: Boolean(externalAccess),
+        userGroups: transformLegacySharingArrayToObject(userGroupAccesses),
+        users: transformLegacySharingArrayToObject(userAccesses),
+    };
+};
+
+export const transformLegacySharingArrayToObject = sharingArray => {
```
Birkbjo, line 94: Remove export

src/i18n/i18n_module_en.properties
```diff
@@ -2361,3 +2361,6 @@ relationship_label=Custom label for relationship
 note_label=Custom label for note
 tracked_entity_attribute_label=Custom label for tracked entity attribute
 sorting_is_disabled_when_filter_is_applied=Sorting is diabled when a filter is applied
+save_the_program_in_order_to_access_sharing_settings=Save the program in order to access sharing settings
+confirm_delete_icon=Are you sure you want to delete this icon?
+intro_icon=Upload, modify and view icons which can be used assigned to other metadata.
```
Birkbjo, line 2366: This was missed when implementing icons. Since this PRs needs to be backported to patch/41, I opted to just sneak it in here 😇

HendrikThePendric (approved):
Hi @Birkbjo, I have only reviewed the code changes and will leave the functional testing to a team-platform engineer. All changes look very sensible to be. Ready to merge as far as I'm concerned.

src/EditModel/event-program/program-access/utils.js
```diff
@@ -1,33 +1,35 @@
...
+    const sharingB = modelB.sharing;
+    if (sharingA.public !== sharingB.public) return false;
+    if (!!sharingA.externalAccess !== !!sharingB.externalAccess) return false;
 
     const compareFunction = (a, b) => a.id < b.id;
     if (
         !isEqual(
-            Array.sort(a.userAccesses || [], compareFunction),
-            Array.sort(b.userAccesses || [], compareFunction),
+            Array.sort(sharingA.users || [], compareFunction),
```
Birkbjo, line 12: I have to double check this. Because this isn't actually an array anymore... I've tested it and it kind of works, but needs investigation... Also where does `Array.sort` come from?

PR #2869, opened 2024-06-03
fix(categoryOptionCombo): remove shortName

Birkbjo:
`shortName` is not a persisted field for `categoryOptionCombo`, so saving this didn't do anything. `categoryOptionCombo` did not have any entry in `field-order.js`, so I've added this. Previously it just got the rendered fields from the schema.


tomzemp (approved):
lgtm 👍

PR #2870, opened 2024-06-03
fix(date-picker): ignore timezone in date-picker

Birkbjo:


tomzemp (approved):
Looks good to me 🎉.

I tested with server in Oslo (CEST) and with browser time set to Bogotá and Vientiane, respectively. Worked as expected. 

I still think it makes more sense to have `substring(0,23)` after `toISOString()` in getISOFormatLocalTimestamp, in order to get rid of the `Z` at the end of the string. 


With this PR we:
- select a date with Material UI date picker. Material UI thinks we want the date at 00:00 local time. If we've selected 31 May and we're in Vientiane, that is then 30 May 17:00 UTC. 
- we apply the conversion function (`(date.getTime() - date.getTimezoneOffset() * 60000).toISOString()`, we get a string back `2024-05-31T00:00:00.000Z`
- we send the date off and the server ignores the `Z` and just assumes that we want `2024-05-31T00:00:00.000` server time

I guess my argument for removing the `Z` is that the `Z` stands for zero offset (from UTC), so with the `Z` we are saying we are sending the time in UTC, but what we want (and what the backend does because it ignores time zones) is to have the date string representation of the time in the server time zone (without any time zone information).

PR #2885, opened 2024-07-02
feat: add option to pick render as tabs direction for data sets

flaminic:
Implements [DHIS2-17506](https://dhis2.atlassian.net/browse/DHIS2-17506).
This PR sets the user dataset decide the direction of a section tabs if the user wants to render as tabs

### Implementation details

The displayOptions was added as a JSON type in the BE to allow us flexibility to add more options without updating the back-end, similarly to what was done for sections. See discussion in https://github.com/dhis2/dhis2-core/pull/16562 

### Background

This is [an RFC](https://dhis2.atlassian.net/wiki/spaces/SOFTWARE/pages/121995265/Form+configuration+RFC) that describes the approach and the priorities for form configuration options. This is based on a [thorough investigation](https://docs.google.com/presentation/d/1pSeFoF91HDYNeL88GghKLvj3wpYf0qUCdhkKXklB2MY/edit#slide=id.g2462c2dd6c9_0_53) by the functional design team for custom form use cases in real-life implementations. Based on that investigation, the ability to show section tabs vertically were one of the main reasons people choose to go the custom forms route so we're tackling these first.

### UI

https://github.com/dhis2/maintenance-app/assets/5418349/ed70bb16-e029-4c2b-9ad8-c22a09f862da

### Related PRs: 

[backend](https://github.com/dhis2/dhis2-core/pull/17891)
[data entry app](https://github.com/dhis2/aggregate-data-entry-app/pull/389)

[DHIS2-17506]: https://dhis2.atlassian.net/browse/DHIS2-17506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/config/field-overrides/dataSet.js
```diff
@@ -1,7 +1,98 @@
...
+        this.props.model.renderAsTabs = renderAsTabs
+        this.updateTabsDirection(tabsDirection)
+    }
+
+
+    render() {
+        const state = this.state;
+        return <div>
+           <Checkbox
+            labelText="Render as tabs"
```
Birkbjo, line 67: Should translate this (and other labels). You can add the d2 context to get access to i18n by wrapping the component with addD2Context:

```
import addD2Context from 'd2-ui/lib/component-helpers/addD2Context';

// ...
   ['renderAsTabs', {
        component: addD2Context(CustomTestComponent)
    }]

// in component:
// this.context.d2.i18n.getTranslation('render_as_tabs')
```

Add keys in `i18n_module_en.properties`.

src/config/field-overrides/dataSet.js
```diff
@@ -1,7 +1,98 @@
...
 import PeriodTypeDropDown from '../../forms/form-fields/period-type-drop-down';
+import Checkbox from '../../forms/form-fields/check-box';
+import {RadioButton, RadioButtonGroup} from "material-ui/RadioButton";
+import log from "loglevel";
+
+class CustomTestComponent extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            renderAsTabs: props.value,
```
Birkbjo, line 14: I don't think it's necessary to add `renderAsTabs` to the state. As long as we call `props.onChange` we should be able to just use the `props.value` for this.

src/config/field-overrides/dataSet.js
```diff
@@ -1,7 +1,98 @@
...
+        this.props.model.displayOptions = JSON.stringify(newDisplayOptions)
+    }
+
+    onDisplayOptionsChanged = (event) =>  {
+        const tabsDirection = event.target.value
+        this.updateTabsDirection(tabsDirection)
+
+    }
+
+    onDisplayAsTabsChanged = (event) =>  {
```
Birkbjo, line 47: Minor: could we rename this to `onRenderAsTabsChanged` to be consistent with the name of the field/property?

src/config/field-overrides/dataSet.js
flaminic: YES. I completely forgot about translations. Thanks!

src/config/field-overrides/dataSet.js
flaminic: good point!

src/config/field-overrides/dataSet.js
flaminic: absolutely

src/config/field-overrides/dataSet.js
```diff
@@ -1,7 +1,98 @@
...
+            return undefined
+        }
+    }
+
+    updateTabsDirection = (tabsDirection) => {
+        const newDisplayOptions = {
+            ...this.state.displayOptions,
+            tabsDirection
+        }
+        this.setState((prevState, _) => ({
```
Birkbjo, line 35: It's actually not necessary to merge in the `prevState` using `this.setState`, it's shallowly merged with the `state`.

Birkbjo (approved):

PR #2894, opened 2024-07-25
fix(dataset/workflow): Add catcombos to workflows and make workflows …

benguaraldi:
The full title was "fix(dataset/workflow): Add catcombos to workflows and make workflows only selectable when their catcombo matches the data set's" but it looks like that was too long.

This code implements all four requests from [DHIS2-17576](https://dhis2.atlassian.net/browse/DHIS2-17576).

[DHIS2-17576]: https://dhis2.atlassian.net/browse/DHIS2-17576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,7 +122,7 @@
     },
     "developer": {
       "url": "",
-      "name": "Mark Polak"
+      "name": "HISP"
     },
     "activities": {
       "dhis": {
--- a/src/config/field-config/field-order.js
+++ b/src/config/field-config/field-order.js
@@ -493,6 +493,7 @@ const fieldOrderByName = new Map([
         'name',
         'periodType',
         'dataApprovalLevels',
+        'categoryCombo',
     ]],
     ['optionGroup', [
         'name',
--- a/src/config/field-overrides/dataApprovalWorkflow.js
+++ b/src/config/field-overrides/dataApprovalWorkflow.js
@@ -4,4 +4,11 @@ export default new Map([
     ['periodType', {
         component: PeriodTypeDropDown,
     }],
+    ['categoryCombo', {
+        referenceType: 'categoryCombo',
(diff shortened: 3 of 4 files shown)


jimgrace (commented):
I'm not a front-end dev, but this code looks right to me. I've watched Ben test it and it seems to cover all the cases presented in the Jira ticket.

benguaraldi:
@Birkbjo and @tomzemp, I started with @jimgrace reviewing to make sure the functionality correct, but now we believe it works properly.  Let me know what you both think of the code.

Thanks!

Birkbjo (commented):
Thanks a lot for implementing this, and very impressive to get into this code!

Just a question though; it seems like the backend will "default" to the `default-catcombo` if null is sent for catcombo in workflows. I guess this is intended for `2.42`?

jimgrace:
@Birkbjo Thanks, good point!

Since approximately 2018, when workflows were created or edited, the backend assigned them the default cat combo. This change to the Maintenance app simply lets a user choose a different cat combo.

Once version 2.42 comes out, the flyway script I'm writing will change workflows with null or default cat combos to the correct cat combo based on their data sets.

As such, we believe that the way the workflow cat combo works in the Maintenance app is acceptable prior to version 2.42 and is the right approach for 2.42 and later.

benguaraldi:
@Birkbjo Thank you for reviewing!  This app is definitely wild to work with.

I defer to @jimgrace on how the workflow cat combos work and should work.

Birkbjo (approved):

benguaraldi:
Thanks, @Birkbjo!

PR #2930, opened 2024-09-04
fix(analyticstablehooks): Adds analytics table hooks to the Maintenance app

benguaraldi:
This adds analytics table hooks to the Maintenance app, as requested in [DHIS2-13473](https://dhis2.atlassian.net/browse/DHIS2-13473).

Note that I alphabetized the order of the items in the Other section and moved Predictors to the same section as Indicators, as the section had gotten too long to show all of the items, at least on my screen.

[DHIS2-13473]: https://dhis2.atlassian.net/browse/DHIS2-13473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Birkbjo:
Thanks a lot for your contributions lately! I will follow up the open PRs shortly, I've been on vacation the last week. 

> moved Predictors to the same section as Indicators, as the section had gotten too long to show all of the items, at least on my screen.

While I agree predictors could be grouped with "indicators", I don't think we should move stuff around in this app, since it can be confusing for users if the location of the section has suddenly changed. 
I took the freedom to push directly to your branch, and move them back, while adding `overflow` to the sidebar. Would be great if you could test if that works for you!

benguaraldi:
@Birkbjo That's a much better solution!  Thank you!  I tested it and it looks good to me.

PR #2961, opened 2024-10-14
feat: add custom text to data sets

flaminic:
Implements [DHIS2-18129](https://dhis2.atlassian.net/browse//DHIS2-18129).
This PR allows users to add a title/ substitle to datasets

## Implementation details
The displayOptions was added as a JSON type in the BE to allow us flexibility to add more options without updating the back-end. See discussion in https://github.com/dhis2/dhis2-core/pull/16562

## Background

This is [an RFC](https://dhis2.atlassian.net/wiki/spaces/SOFTWARE/pages/121995265/Form+configuration+RFC) that describes the approach and the priorities for form configuration options. This is based on a [thorough investigation](https://docs.google.com/presentation/d/1pSeFoF91HDYNeL88GghKLvj3wpYf0qUCdhkKXklB2MY/edit#slide=id.g2462c2dd6c9_0_53) by the functional design team for custom form use cases in real-life implementations. Based on that investigation, the ability to add custom text to sections were one of the main reasons people choose to go the custom forms route so we're tackling these first.

## Preview
https://github.com/user-attachments/assets/94906d06-c531-416b-971e-6912e74d3eaa


src/config/field-overrides/dataSet.js
```diff
@@ -52,39 +54,149 @@ class RenderAsTabsSettings extends React.Component {
...
+                value={state.displayOptions && state.displayOptions.customText !== undefined}
+                onChange={this.onAddCustomTextChanged}
+            />
+
+            {state.displayOptions && state.displayOptions.customText &&
+                <RadioButtonGroup
+                    onChange={this.onCustomTextAlignmentChanged}
+                    name="customTextAlignment"
+                    defaultSelected={
+                        (state.displayOptions && state.displayOptions.customText && state.displayOptions.customText.align) || 'left'}
```
tomzemp, line 160: do you want `left` or `center` to be the default? It seems with `onAddCustomTextChanged`, that the alignment was being set to center by default?

src/config/field-overrides/dataSet.js
```diff
@@ -52,39 +54,149 @@ class RenderAsTabsSettings extends React.Component {
...
+                        value='right'
+                        label={this.translate('right')}
+                        style={{margin: '10px'}}
+                    />
+                </RadioButtonGroup>}
+
+                {state.displayOptions && state.displayOptions.customText &&
+                <div style={customTextCssStyles}>
+                    <TextField
+                    value={(this.state.displayOptions && state.displayOptions.customText &&
```
tomzemp, line 185: is there a reason to use `this.state` vs `state` (applies a couple of other places within the render too). It seems like `const state = this.state` is declared above

src/config/field-overrides/dataSet.js
```diff
@@ -52,39 +54,149 @@ class RenderAsTabsSettings extends React.Component {
...
+                : undefined
+
+        const newDisplayOptions = {
+            ...this.state.displayOptions,
+            customText
+        }
+        this.updateDisplayOption(newDisplayOptions)
+    }
+
+    onCustomTextAlignmentChanged = (event) =>  {
```
tomzemp, line 78: You could also combine these `onCustomTextAlignmentChanged`, `onCustomTextHeaderChanged`, `onCustomTextSubheaderChanged`, by doing something like

```
    onCustomTextChanged = (event, field) =>  {
        const customText =
            {...this.state.displayOptions[field],  [field]: event.target.value}

        const newDisplayOptions = {
            ...this.state.displayOptions,
            customText
        }
        this.updateDisplayOption(newDisplayOptions)
    }
```
but I'm not sure if the reused code makes up for the redice readability

tomzemp (approved):
Looks good. I've left a couple of small comments.

src/config/field-overrides/dataSet.js
flaminic: THANK YOUUUUU

src/config/field-overrides/dataSet.js
flaminic: ehehe tiredness ;) changed

src/config/field-overrides/dataSet.js
flaminic: went for an hybrid between the 2. let me know what you think!

PR #3064, opened 2025-03-24
fix(trackedEntityType): add shortName and code fields

Birkbjo:


Birkbjo:
Closed in favor of #3065

PR #3069, opened 2025-04-09
fix: correctly list manifest shortcuts as an array instead of an object

amcgee:
This issue was caused because package patches were not being applied - the yarn.lock hash wasn't updated, so the cache had a stale copy of the dependencies which should have been patched.  This ended up breaking the dhis2-core build


amcgee:
This needs to be a `fix` to trigger a new build/release since the artifact created in the previous release had the issue (see [here](https://github.com/d2-ci/maintenance-app/blob/master/manifest.webapp#L21-L24))

amcgee:
Corrected manifest after this fix: https://github.com/d2-ci/maintenance-app/blob/fix-gh-action-build-cache/manifest.webapp

PR #3070, opened 2025-04-09
feat(manifest): add shortcuts for command palette [DHIS2-19411] (42.0)

amcgee:
Backport of #3067 and #3069 

Approved by RCB

Question - since this is one of very few apps which still use a branching/backporting strategy, we need to choose a version number that makes sense here.... there is exactly one change on master that's not one this branch (https://github.com/dhis2/maintenance-app/pull/3059 @flaminic) which means that the version number on this branch is no longer correct.  Maybe we should post-fix it with `-42.0` or something so that the app version is unique at least, if not semantically correct?

ALTERNATIVE: instead of backporting here, we could just approve the very small change referenced above and then point to the latest built version tag on master from the dhis2-core apps-to-bundle.json (cc @Philip-Larsen-Donnelly)

### DIFF
--- a/.github/workflows/dhis2-verify-app.yml
+++ b/.github/workflows/dhis2-verify-app.yml
@@ -30,7 +30,7 @@ jobs:
               id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
               with:
                   path: '**/node_modules'
-                  key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+                  key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock', 'patches/**') }}
 
             - name: Install
               if: steps.yarn-cache.outputs.cache-hit != 'true'
@@ -50,7 +50,7 @@ jobs:
               id: yarn-cache
               with:
                   path: '**/node_modules'
-                  key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+                  key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock', 'patches/**') }}
 
             - name: Build
               run: yarn build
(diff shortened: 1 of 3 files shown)


Birkbjo:
> Question - since this is one of very few apps which still use a branching/backporting strategy, we need to choose a version number that makes sense here.... there is exactly one change on master that's not one this branch (https://github.com/dhis2/maintenance-app/pull/3059 @flaminic) which means that the version number on this branch is no longer correct. Maybe we should post-fix it with -42.0 or something so that the app version is unique at least, if not semantically correct?

Since this app is not continuously released or built to be upgraded without backend, I don't think the version number is really significant. At least we haven't really "cared" much about it in regards to backports previously. 

I'm pretty sure that the version number is not bumped in branches that are not master. So when we backport we actually don't bump any version? So I'm not sure if we need to "care" about this? It can be confusing if looking at the version-number when reporting bugs etc, but again - we haven't done anything in regards to this before. 

> ALTERNATIVE: instead of backporting here, we could just approve the very small change referenced above and then point to the latest built version tag on master from the dhis2-core apps-to-bundle.json 

Would this potentially have consequences for future patch releases? Or is the `apps-to-bundle` regenerated during release, and just looks at the state of "backport"-branches at the time of freeze?

amcgee:
@Birkbjo yeah we didn't care about version numbers at all before, but now all apps (even patch-branched ones like maintenance) will list the app's version number more prominently in the profile menu.  What this means is that the version number listed in 2.42.0 weight be identical to the one listed in 2.42.1, but the actual code/artifact would be different.  I think using semantic release to version these apps - even though they're not continuously released - makes a lot of sense.

And you're correct on your last point, the reference in apps-to-bundle.json is "frozen"/updated when we freeze a release.  For both master and 2.42 branches it points to the app's master branch currently, whereas on the patch/2.42.0 branch of core it points to the patch/2.42.0 branch here.  I'm just advocating for pointing it at the semantic release-generated version tag (which could be on master or on a release branch live `v30.2.x`) instead of to a branch which isn't truly "pinned" and has misleading version numbers.

amcgee:
Not a gigantic issue either way here, we can just do this back port if we want, but we could address it now if we want

Birkbjo (approved):

PR #3089, opened 2025-09-03
fix: update program indicators orgunit descriptions

tomzemp:
Updates the text for program indicator organisation unit descriptions:

image

Note, I've updated the non-English translation files. Usually, we would not do that, but the setup for the maintenance app does not fall back to English and instead uses the value from each translation file, so I thought it made more sense to just do the full copy/replace.


Sharmyn28 (approved):
LGTM! :smile:

